PR #1, opened 2019-08-07
adding package.json

jonathanburns:


package.json
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "@lyft/flyteidl",
+    "version": "0.1.0",
+    "description": "Compiled protocol buffers and gRPC service clients/servers for Flyte IDLs",
+    "repository": {
+        "type": "git",
+        "url": "git@github.com:lyft/flyteidl"
+    },
+    "author": "Randy Schott <rschott@lyft.com>",
```
dschaller: maybe we should use a team distro-list here?

schottra: We don't have a good public-facing list for Flyte yet. And the convention in our Python packages seems to be using a single person as the author. I'm okay being that person until we figure out the communications. We can always change it later.

PR #8, opened 2019-08-30
Add admin service endpoint documentation

katrogan:


protos/flyteidl/service/admin.proto
```diff
@@ -155,13 +151,19 @@ service AdminService {
...
+    };
   }
 
   rpc UpdateLaunchPlan (flyteidl.admin.LaunchPlanUpdateRequest) returns (flyteidl.admin.LaunchPlanUpdateResponse) {
     option (google.api.http) = {
       put: "/api/v1/launch_plans/{id.project}/{id.domain}/{id.name}/{id.version}"
       body: "*"
     };
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+      description: "Update the status of an existing launch plan definition."
```
matthewphsmith: Since this is _really_ the core of the entire platform, I think this is a natural space to explain the semantics of the platform.  Most endpoints are pretty self-explanatory so they don't need much text, but for things like update launch plan, it would be good to describe what setting a launch plan to active means.

Also, describing the idempotent behavior of our create endpoints would be helpful.

katrogan: sure, done for both

protos/flyteidl/service/admin.proto
```diff
@@ -155,13 +187,22 @@ service AdminService {
...
   rpc UpdateLaunchPlan (flyteidl.admin.LaunchPlanUpdateRequest) returns (flyteidl.admin.LaunchPlanUpdateResponse) {
     option (google.api.http) = {
       put: "/api/v1/launch_plans/{id.project}/{id.domain}/{id.name}/{id.version}"
       body: "*"
     };
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+      description: "Update the status of an existing launch plan definition. "
+      "At most one launch plan version for a given {project, domain, name} can be active at a time. "
+      "If this call sets a launch plan to active and existing version is already active, the result of this call will be that the "
+      "formerly active launch plan will be made inactive and specified launch plan in this request will be made active."
```
matthewphsmith: We should mention that if the launch plan is associated with a schedule, then the schedule will be activated while the existing schedule will be deactivated.

protos/flyteidl/service/admin.proto
```diff
@@ -65,18 +76,18 @@ service AdminService {
...
+      responses: {	
+        key: "400"	
+        value: {	
+          description: "Returned for bad request that may have failed validation."	
+        }	
+      }	
+      responses: {	
+        key: "409"	
+        value: {	
+          description: "Returned for a request that references an identical entity that has already been registered."
```
matthewphsmith: I don't think this is really descriptive of our idempotent behavior. It isn't returned for an identical entity, it is returned for an entity that has already been registered with the same ID _and_ which is not identical to the original entry.

katrogan: the translation from grpc status code to http is confusing, but this is in fact what we return when the digests match (https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/errors.go#L31). we return InvalidArgument otherwise but there's a lot of scenarios where InvalidArgument is the response so i don't think calling all of them out is particularly useful since they're likely to evolve

protos/flyteidl/service/admin.proto
katrogan: ok, done

protos/flyteidl/service/admin.proto
matthewphsmith: ahh i see

PR #10, opened 2019-09-05
default-stackdriver-user

vsbus:
We had `&authuser=1` argument in the stackdriver links that was opening with second logged-in user. While some users logged in to a personal gmail first, many other in the organization are logged in with their lyft accounts first.
It is better to not specify this argument and expect all L5 Flyte users are logged in to gmail with @lyft.com account first. If that isn't the case we will instruct them to sign out from all google accounts and sign in again with @lyft.com first.


vsbus:
PTAL @kumare3 
CC @sumedhsakdeo

vsbus:
PTAL @migueltol22 @wzpher

vsbus:
🚀

PR #12, opened 2019-10-01
Adding a new Workflow execution phase LAUNCHED

anandswaminathan:
cc @schottra @katrogan @kumare3 

Not sure if this would break UI. Will have to check.


protos/flyteidl/core/execution.proto
```diff
@@ -18,6 +18,7 @@ message WorkflowExecution {
         FAILED = 6;
         ABORTED = 7;
         TIMED_OUT = 8;
+        LAUNCHED = 9;
```
katrogan: Should we added a comment about what this means since it's not immediately clear how this differs from QUEUED?

wild-endeavor:
what's the point of this though?  can we document the difference between queued, launched, and running?

wild-endeavor:
what she said.

protos/flyteidl/core/execution.proto
anandswaminathan: Actually 'QUEUED' might be a better name. cc @kumare3

anandswaminathan:
@wild-endeavor @katrogan Closing this PR. `QUEUED` does the job for me. I thought it was used for something else. But looks like it is not.

PR #13, opened 2019-10-08
Deprecate Literal Map in Execution Closure and Execution Spec

matthewphsmith:
Placing the literal map of inputs directly in the spec and closure leads to bloated messages when not necessary.  This PR deprecates that in favor of a pattern that is used elsewhere (node and task executions) to fetch inputs/outputs separately.


protos/flyteidl/admin/execution.proto
```diff
@@ -26,6 +26,10 @@ message ExecutionCreateRequest {
 
     // Additional fields necessary to launch the execution.
     ExecutionSpec spec  = 4;
+
+    // The inputs required to start the execution. All required inputs must be
+    // included in this map. If not required and not provided, defaults apply.
+    core.LiteralMap inputs   = 5;
```
katrogan: why move the inputs out of the executionspec too?

schottra: I would also be hesitant about doing this (it will break launch support until UI/CLI are updated). I guess the input size issue _could_ still be a problem if the default values are large. Does that seem like a common occurrence?

matthewphsmith: The execution spec is returned in the Execution too so unless I just leave that empty by convention, the message will be very large.  I could add a UrlBlob in the Execution too for inputs and outputs, but that would be unique compared to what we do for task and node executions.

PR #15, opened 2019-10-09
Workflow ID comment states that id is autogenerated, that's no longer…

jonathanburns:
… true.


katrogan:
I think you might need to regen however since the comments are added to generated code too

PR #17, opened 2019-10-24
add sns message proto

migueltol22:
Adding SNS Message proto to be able to send notification with additional common data to be used by different topics.


kumare3:
Can we close this PR out?

migueltol22:
Closing as won't do.

PR #19, opened 2019-10-24
Upgrade to Go 1.13

EngHabu:


Gopkg.toml
```diff
@@ -57,6 +60,18 @@ required = ["github.com/envoyproxy/protoc-gen-validate"]
   name = "google.golang.org/grpc"
   version = "1.16.0"
 
+[[constraint]]
+  name = "github.com/vektra/mockery"
+  source = "github.com/enghabu/mockery"
```
schottra: Did you mean to leave this in?

EngHabu: Yes... until it gets merged to master... it's only used to generate.. no "production dependency" per se

PR #20, opened 2019-10-25
Identifier descriptions

schottra:
* Updated docker image used for `protocgenerator`. We need this latest version to allow enum types as URL parameters
* Added new type `NamedEntity`. This is an extension of our previous concept of `NamedEntityIdentifier`. For now, it composes an id and a metadata object
* Added a new type `NamedEntityMetadata`
* Added a list endpoint for fetching `NamedEntity` objects for a type/project/domain combo. Unlike our previous implementations, this one allows handling all types in one endpoint. We were previously limited by grpc-gateway not allowing enum values as URL path parameters.
* Added get/update endpoints for `NamedEntity`. This is to support fetching/updating the metadata associated with that object in cases where you know the `id` portion of the message.


protos/flyteidl/admin/common.proto
```diff
@@ -7,18 +7,26 @@ import "flyteidl/core/execution.proto";
...
     // A domain can be considered as a subset within a specific project.
-    string domain     = 2;  // [(validate.rules).string.min_bytes = 1];
+    string domain                = 2;  // [(validate.rules).string.min_bytes = 1];
     // User provided value for the resource.
     // The combination of project + domain + name uniquely identifies the resource.
     // +optional - in certain contexts - like 'List API', 'Launch plans'
-    string name       = 3;
+    string name                  = 3;
+
+    NamedEntityMetadata metadata = 4;
```
katrogan: why is the metadata part of the identifier?  can you not create a separate message encapsulates the identifier and the metadata?

schottra: The intention is that you should be able to list the identifiers and also have a description of what each identifier is.

schottra: I moved the metadata out and used a message composing id and metadata, and made that the preferred type when returning a list of identifiers. ptal!

protos/flyteidl/admin/common.proto
```diff
@@ -25,8 +19,18 @@ message NamedEntityIdentifier {
...
+
+message NamedEntityMetadata {
+    // Common description across all versions of the entity
+    // +optional
+    string description = 1;
+}
 
-    NamedEntityMetadata metadata = 4;
+// Composes a NamedEntityIdentifier with its associated metadata
+message NamedEntityIdentifierWithMetadata {
```
katrogan: thank you for refactoring! since this name is a bit long, what do you think about just "NamedEntity" ?

schottra: I agree it's long. But since the original message is specifically called an "Identifier", I'm hesitant to remove that word in an effort to keep it consistent 😬

katrogan: sure, but the original message only handled identifier entities :). elsewhere we just refer to the underlying entity in message names rather than do something like ExecutionWithMetadata or WorkflowWithClosure

katrogan: I see what you're saying with the existing endpoint being named `NamedEntityIdentifierListRequest`. Perhaps, rather than deprecating the field within the endpoint we deprecate the endpoint altogether and just add a new one that better describes what this endpoint now returns like `NamedEntityListRequest`

schottra:
@katrogan I reworked it so that
* We have a new `NamedEntity` message, which includes an `id` and a `metadata` field. This can be expanded in the future
* We have a new request/response type for listing
* We have a new endpoint for listing, which uses an enum value as a path parameter. So we only need one endpoint to handle listing entities for any type
* I updated my metadata endpoints to be a get/update for a NamedEntity. This allows us to update additional fields with the same request later if needed. (I don't plan to allow modification of the id or resource type)

Please give it one more look and let me know if anything seems strange. Thanks!

PR #21, opened 2019-10-28
Add project domain attributes proto & corresponding update endpoint

katrogan:
Naming is hard so please let me know if you have better suggestions on what to call this message!


protos/flyteidl/admin/project_domain_attributes.proto
```diff
@@ -0,0 +1,26 @@
...
+
+message ProjectDomainAttributes {
+    // Unique project id for which this set of attributes will be applied.
+    string project                 = 1;
+
+    // Unique domain id for which this set of attributes will be applied.
+    string domain                  = 2;
+
+    // Custom resource attributes which will be applied in cluster resource creation (e.g. quotas).
+    // Map keys are the names of variables in templatized resource files.
```
EngHabu: case sensitive or not? maybe call that out here...

protos/flyteidl/service/admin.proto
```diff
@@ -376,4 +377,14 @@ service AdminService {
       description: "Retrieve input and output data from an existing task execution."
     };
   }
+
+  rpc UpdateProjectDomainAttributes (flyteidl.admin.ProjectDomainAttributesUpdateRequest) returns (flyteidl.admin.ProjectDomainAttributesUpdateResponse) {
```
EngHabu: Do we have Create? or is Update == CreateOrUpdate? is that common?

schottra: fwiw, In another PR for entity metadata, I'm following a create-or-update pattern with an UPDATE endpoint

katrogan: @EngHabu for this i didn't want to bother with create or update since the config'd defaults will always be applied and i didn't want to have to expose the underlying model to the end user. this does depart from the explicit create in our other models but since this has a default anyways i think it makes sense

protos/flyteidl/admin/project_domain_attributes.proto
katrogan: good point, done

PR #24, opened 2019-11-12
More resource attributes and applicable scopes

katrogan:


protos/flyteidl/admin/project_attributes.proto
```diff
@@ -0,0 +1,27 @@
...
+option go_package = "github.com/lyft/flyteidl/gen/pb-go/flyteidl/admin";
+
+message ProjectAttributes {
+    // Unique project id for which this set of attributes will be applied.
+    string project                 = 1;
+
+    // Custom resource attributes which will be applied in cluster resource creation (e.g. quotas).
+    // Map keys are the *case-sensitive* names of variables in templatized resource files.
+    // Map values should be the custom values which get substituted during resource creation.
+    map<string, string> attributes = 2;
```
kumare3: WE should also add REsource manager quotas in this

katrogan: That's what these are already being used for :) (in project_domain_attributes.proto)

katrogan:
ping @EngHabu - refactored this a bit

PR #26, opened 2019-11-29
generated java service stub

honnix:
This is generated by slight modifying `lyft/protocgenerator` image.


honnix:
OK this doesn't work. It would be nice if you can open source the image.

PR #27, opened 2019-11-29
migrate to go module

honnix:


honnix:
Might relate to https://github.com/golang/go/issues/27455 and https://github.com/golang/go/issues/30185

honnix:
I think merge https://github.com/lyft/flytestdlib/pull/50 and release a new `pflags` will get this fixed.

honnix:
@@EngHabu Could you please take a look at this as well? Thanks.

honnix:
I compiled `pflags` and verified locally.

honnix:
https://github.com/golang/go/issues/30515

honnix:
https://github.com/golang/go/issues/33008

generate_mocks.sh
```diff
@@ -1,7 +1,13 @@
...
-which mockery || (go install github.com/lyft/flyteidl/vendor/github.com/vektra/mockery/cmd/mockery)
+which mockery || (go get github.com/vektra/mockery/cmd/mockery)
 
 mockery -dir=gen/pb-go/flyteidl/service/ -name=AdminServiceClient -output=clients/go/admin/mocks
 mockery -dir=gen/pb-go/flyteidl/datacatalog/ -name=ArtifactsClient -output=clients/go/datacatalog/mocks
+
+# workaround https://github.com/golang/go/issues/33008
+# here we can not install outside of current module because we need the replacement in
+# go.mod
+rm go.sum
```
wild-endeavor: Read the linked issue, but I'm still a little bit confused.  Does running the steps above in this shell script add something to the `go.sum` file that shouldn't be there?  If so, I think I'd feel more comfortable if we just copied the go.sum file to a tmp file before doing anything, and then moving it back afterwards.  The reason is that i just don't want to rely on `go mod tidy` to correctly produce the exact same file.

What do you think?

cc @EngHabu

EngHabu: I agree. Let's do that, and add a catch clause to guarantee it's copied back if this script is interrupted in anyway...

wild-endeavor: Haytham brings up a good point.  Could we make the copying of it back a bash `trap`?  Just to ensure that if running mockery fails for some reason, it's still set back.

honnix: Makes sense. Will take a look.

honnix: This might be simplified by copying go.mod and go.sum to a tmp folder and installing mockery from there. So we won't need a trap.

honnix:
Don't understand why it adds all those `h1` lines to go.sum. :(

honnix:
LGTM now but I cannot approve myself. :)

PR #32, opened 2019-12-31
Add get and delete methods for attributes

katrogan:
Currently ProjectAttributes, ProjectDomainAttributes and WorkflowAttributes can be used to specify resource overrides. The only existing endpoint is to update (or implicitly create) these overrides but the delete semantics are undefined. To remedy this, this change introduces a delete API for all these customizable attributes AND a get to check what the existing values are currently set as.

In the future, only system administrators will be able to mutate these values once we have authorization rolled out.


wild-endeavor:
Can you add a better description?

protos/flyteidl/admin/project_domain_attributes.proto
```diff
@@ -23,3 +23,29 @@ message ProjectDomainAttributesUpdateRequest {
 // Purposefully empty, may be populated in the future.
 message ProjectDomainAttributesUpdateResponse {
 }
+
+message ProjectDomainAttributesGetRequest {
```
EngHabu: We should either allow people to select the resource type (the one of thing in MatchingAttibute) or return a list of attributes... right?

katrogan: wow yes, thank you for the catch!

katrogan: done

katrogan:
@wild-endeavor 

> Can you add a better description?

Done.

katrogan:
friendly ping @wild-endeavor @EngHabu

PR #33, opened 2020-01-03
[PoC] migrate to go module, alternative

honnix:
Alternative to https://github.com/lyft/flyteidl/pull/27


honnix:
@wild-endeavor PTAL

honnix:
Downside of this approach is one can not run `go mod tidy` under `boilerplate/lyft/golang_support_tools/`, because that removes everything.

PR #34, opened 2020-01-10
Remove endpoint for Project Attributes

anandswaminathan:


katrogan:
Delete https://github.com/lyft/flyteidl/blob/master/protos/flyteidl/admin/project_attributes.proto as well?

anandswaminathan:
@katrogan Done

PR #38, opened 2020-02-07
[Placement Manager] Add override ability for running a particular execution

anandswaminathan:
We want the ability to add overrides to route executions to a particular cluster. 
This will follow the similar hierarchical rules when selecting overrides. 

Example: 
Workflow-A can be routed to cluster Label A.
Domain-B can be routed to cluster Label B. 

cc @katrogan


protos/flyteidl/admin/matchable_resource.proto
```diff
@@ -52,5 +55,7 @@ message MatchingAttributes {
     ClusterResourceAttributes cluster_resource_attributes = 2; 
 
     ExecutionQueueAttributes execution_queue_attributes = 3;
+
+    string cluster_label = 4;
```
katrogan: do we ever anticipate this changing? is it worth creating a separate proto so we can future-proof this attribute type if we ever want to add or update the way we specify a cluster? (e.g. a list of labels)

anandswaminathan: I explicitly wanted to make sure we do not need a list.

But I see benefit of a separate proto. Will create one.

protos/flyteidl/admin/matchable_resource.proto
```diff
@@ -44,6 +47,11 @@ message ExecutionQueueAttributes {
   repeated string tags           = 1; 
 }
 
+message ExecutionCluster {
```
akhurana001: calling this a cluster doesn't seem right to me

anandswaminathan: We discussed offline, and agreed on ExecutionCluster.
cc @surindersinghp

akhurana001:
Looks good

akhurana001:
👍

PR #41, opened 2020-02-20
Add interruptible field 

migueltol22:


protos/flyteidl/admin/execution.proto
```diff
@@ -222,6 +222,9 @@ message ExecutionSpec {
 
     // Annotations to apply to the execution resource.
     Annotations annotations   = 8;
+
+    // Identify whether execution is interruptible
+    bool interruptible = 9;
```
surindersinghp: can we move it under execution metadata? https://github.com/lyft/flyteidl/pull/41/files#diff-a985470721c92a43bf4f75803828617bR205

migueltol22: done.

PR #42, opened 2020-02-24
Add iterruptible field

migueltol22:
Add interruptible as part of workflow and node metadata to specify whether the node and/or workflow can be interrupted.


surindersinghp:
@EngHabu

protos/flyteidl/core/workflow.proto
```diff
@@ -124,6 +127,8 @@ message Node {
 // Metadata for the entire workflow.
 // To be used in the future.
 message WorkflowMetadata {
+    // Identify whether workflow is interruptible
```
EngHabu: A couple of things:
- In the generated golang proto, .GetInterruptible() will return false whether it's explicitly set to false or not set at all. Knowing this, how will this flag interact with the per node flag?
- After resolving this ^, please add to the comment here how precedence will work...

protos/flyteidl/core/workflow.proto
```diff
@@ -73,6 +73,9 @@ message NodeMetadata {
 
     // Number of retries per task.
     RetryStrategy retries = 5;
+
+    // Identify whether node is interruptible
+    bool interruptible = 6;
```
EngHabu: I was going to ask you to add as similar flag to message but then I realized you already did:
https://github.com/lyft/flyteidl/blob/master/protos/flyteidl/admin/execution.proto#L191

I think it's important to think about precedence and how will these three work together.... please see my other comment for more details

protos/flyteidl/core/workflow.proto
migueltol22: So the behavior should be whatever value is set at the workflow level than all nodes should be set to that value as well unless a value has already been set at the node level. In this case I believe using a `oneof` value should be the correct message type instead of bool. @EngHabu would you agree?

surindersinghp: We still need a default behavior for the workflow. What do we do if interruptible isn't defined? Yes, the node will take precedence.

migueltol22: @surindersinghp I'd propose for the moment if not set than we should default to **not** interruptible

surindersinghp: then bool should be good, right? default=false.

migueltol22: @surindersinghp if my understanding is correct bool will not work because even if the value is not set .GetInterruptible() will return false. If we have the value set to true at the workflow level there isn't a way to distinguish between a user setting interruptible to false or that value not being set at all at the node level.
Using a `oneof` we can check if the value is not set at the node level then use the value set at the workflow level (with default being false if it isn't set).

surindersinghp: sure, I meant at the workflow level (WorkflowMetadata). Yes, for node level it should be oneof

surindersinghp:
looks good to me. @EngHabu: can you please review as well?

protos/flyteidl/admin/node_execution.proto
```diff
@@ -79,7 +79,10 @@ message NodeExecution {
 
 // Represents additional attributes related to a Node Execution
 message NodeExecutionMetaData {
-    bool interruptible = 1;
+    // Identify whether node is interruptible
```
EngHabu: hmm what is this for?

protos/flyteidl/core/workflow.proto
```diff
@@ -124,6 +129,10 @@ message Node {
 // Metadata for the entire workflow.
 // To be used in the future.
 message WorkflowMetadata {
+    // Identify whether workflow is interruptible.
+    // The value set at the workflow level will be the defualt value used for nodes
```
EngHabu: So the value set at the execution level will trump both, right?

migueltol22:
@EngHabu Updated based off our offline discussion. I didn't cut a release so removing the field on the admin protos should be fine. 
As far as precedence I believe it should be:
1. If value is set at task level, that will take precedence.
2. If not set at task level, value at node level will take precedence.
3. If not set at node level, value at workflow will take precedence, with this value set to false by default. 
@surindersinghp thoughts?

EngHabu:
Thanks @migueltol22. To keep things consistent with Resources (CPU/Mem) I would say we should do the reverse.
1. If the value is set (true/false) at the Execution level, use that, break
2. If the value is set (true/false) at the WF level, use that, break
3. If the value is set (true/false) at the Node level, use that, break
4. If the value is set (true/false) at the Task level, use that, break
5. Else, use default (what's the default?), for backcompat, the default should probably be false? but then I think we sent mixed messages when we announced the change to Development domain. What do you think @surindersinghp @akhurana001 ?

protos/flyteidl/core/workflow.proto
```diff
@@ -124,6 +129,10 @@ message Node {
...
+    // unless explicitly set at the node level.
+    bool interruptible = 1;
```
EngHabu: Can you rename it to reflect it's the default value? or add a WorkflowMetadataDefaults message and reference it here...
CC: @surindersinghp Isn't that what we talked about?

surindersinghp: yes, fixing.

surindersinghp: @EngHabu changing type name shouldn't break anything.

EngHabu: Hmmm 🤞I hope that doesn't break the models code in the SDK

PR #45, opened 2020-03-04
Add Presto query message

lu4nm3:
This message is to support functionality for a Presto task


protos/flyteidl/plugins/presto.proto
```diff
@@ -0,0 +1,16 @@
...
+package flyteidl.plugins;
+
+import "flyteidl/core/tasks.proto";
+
+option go_package = "github.com/lyft/flyteidl/gen/pb-go/flyteidl/plugins";
+
+// This message works with the 'presto' task type in the SDK and is the object that will be in the 'custom' field
+// of a Presto task's TaskTemplate
+message PrestoQuery {
+    string routing_group = 1;
```
wild-endeavor: is there a concept of tags like qubole has for hive?  are there any other metadata fields that our installation of presto has that would be helpful to include?

wild-endeavor: if not sure, +1

wild-endeavor: oh but can you bump the `setup.py` version so that this can be published to pypi?

lu4nm3: @wild-endeavor So there are a couple of ways that you can pass identifying information to  Presto, but it's not through tags. First, there is the `user` which I was planning on deriving from a user's email if they log into Flyte. Then, you can also specify `source` which in this case will be `flyte` to indicate where the query is coming from. Here is a screenshot of what these fields look like from the Presto console for a particular query.
Screen Shot 2020-03-05 at 12 01 19 AM So to answer your original question, I think these should be a good set of starting fields. Should we need to add more in the future, I don't think it would be too much work.

Also, what version should I bump `setup.py` to? Currently it looks like it's set to `0.17.7`. Can I just set it to `0.17.8`?

PR #46, opened 2020-03-06
Add interruptible field to TaskMetadata

migueltol22:
For interruptible we will populate it at the node level but require it be part of TaskMetadata for a user to set the value.

We are using `oneof` instead of `bool` because otherwise we would be unable to distinguish between value being set by the user or defaulting to false.

The logic of handling precedence will be done as part of flytepropeller.


surindersinghp:
looks good to me. @matthewphsmith: can you please take a look as well?

PR #50, opened 2020-03-26
Add max_wait_time field to Workflow and Node

migueltol22:
# TL;DR
Adding max_wait_time to workflow metadata and node metadata. This field will be the max amount of time that a workflow can be delayed due to queueing. 

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


protos/flyteidl/core/workflow.proto
```diff
@@ -78,6 +78,9 @@ message NodeMetadata {
     oneof interruptible_value {
         bool interruptible = 6;
     };
+
+    // Total wait time a workflow can be delayed by queueing.
+    string max_queue_time = 7;
```
EngHabu: Can we use [duration](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/duration.proto) instead?

protos/flyteidl/core/workflow.proto
```diff
@@ -129,7 +132,8 @@ message Node {
 // Metadata for the entire workflow.
 // To be used in the future.
 message WorkflowMetadata {
-    // Default Workflow Metadata for the entire workflow.
+    // Total wait time a workflow can be delayed by queueing.
```
EngHabu: What happens if a node has max wait time of 5 mins while the workflow has 1 min... does that mean we will only allow it to wait for 1min?
If there are "invalid" settings (e.g. WF queue can't be smaller than any node's queue) please update the [compiler](https://github.com/lyft/flytepropeller/tree/master/pkg/compiler) to validate and error on that.

Please also update the docs here to describe how the two interact

migueltol22: In this case, I plan on following the same precedence as Resources. So in your example given Workflow would take precedence. Thanks for the link I'll update the compiler to validate!

surindersinghp: Sorry, there seems to be some confusion:
 Users will not be defining wait-time at the node level, they only provide wait-time for the workflow. From the wf max-wait-time (or queueing-budget, I like this term more :)), we will figure out wait-time we can afford for the nodes in the WF at runtime. Lets do a quick chat tomorrow to be on the same page.

migueltol22: Updated to only be specified in wf metadata. Have no preference over `max_wait_time` or `queueing_budget`

protos/flyteidl/core/workflow.proto
```diff
@@ -129,7 +130,8 @@ message Node {
...
+    google.protobuf.Duration max_wait_time = 1;
```
EngHabu: I like this more too...
```suggestion
    google.protobuf.Duration queuing_budget = 1;
```

surindersinghp:
@migueltol22  Thank you!

PR #55, opened 2020-04-03
Add state to NamedEntity

katrogan:
# TL;DR
This change moves the state concept from individual workflows to named entities instead.

This change is backwards incompatible but reverts an endpoint that no one is calling in production.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
The console uses the list NamedEntityIdentifier endpoint rather than workflow one, so this should be updated for controlling visibility.

## Tracking Issue
https://github.com/lyft/flyte/issues/235

## Follow-up issue
_NA_
OR


protos/flyteidl/admin/common.proto
```diff
@@ -21,10 +21,23 @@ message NamedEntityIdentifier {
     string name                  = 3;
 }
 
+// The status of the named entity is used to control its visibility in the UI.
+enum NamedEntityState {
+    // By default, all workflows are considered active and under development.
```
schottra, line 26: nit on this comment being specific to workflows

katrogan: done

PR #58, opened 2020-04-16
Adding idempotent recorder that emits only unique events

surindersinghp:
This convenience layer only sends event if there is a phase change or the last event was sent long ago (more than maxupdatetime). The parameter maxupdatetime is configurable with default value of 5mins

# TL;DR

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


clients/go/events/idempotent_recorder.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+type idempotentWorkFlowEventRecorder struct {
+	internalRecorder    *eventRecorder
+	events              map[string]EmittedEventInfo
+	maxUpdateLagSeconds int64
+	mutex               *sync.Mutex
+}
+
+func (i *idempotentWorkFlowEventRecorder) idempotentRecord(ctx context.Context, id string, e proto.Message, phase string, terminal bool) error {
```
kumare3, line 30: this wouldnt work right, when will you remove the events, Why not use LRU map that we have?

surindersinghp: we remove it here: https://github.com/lyft/flyteidl/pull/58/files#diff-88bad519713daa3e85d2e82c787119ccR47

basically on an terminal event. I actually had an LRU map for further keep the size in check for the case when really large number of WFs are running :) I removed it as it seemed like an unnecessary complication for our scenario (with explicit terminal state)

clients/go/events/idempotent_recorder.go
```diff
@@ -0,0 +1,99 @@
...
+		if err != nil {
+			if !eventsErr.IsAlreadyExists(err) {
+				return err
+			}
+			logger.Infof(ctx, "Workflow event phase: %s, executionId %s already exist", phase, id)
+		}
+
+		// delete the event from events map if it is an terminal event
+		if ok && terminal {
+			delete(i.events, id)
```
kumare3, line 47: but then what if we send multiple terminal events? I still feel rather than having our own global lock, i think i prefer LRU cache? it also allows the map to now grow unbounded?

surindersinghp: terminal events will be emitted every time without going through the cache (first terminal event will remove the entry from cache).

OK, I  add a LRU cache.

surindersinghp:
@anandswaminathan @EngHabu @kumare3 Gentle Ping!

clients/go/events/idempotent_recorder.go
```diff
@@ -0,0 +1,102 @@
...
+
+func (i *idempotentWorkFlowEventRecorder) RecordNodeEvent(ctx context.Context, e *event.NodeExecutionEvent) error {
+	id := e.Id.String()
+	terminal := e.Phase == core.NodeExecution_SUCCEEDED || e.Phase == core.NodeExecution_FAILED ||
+		e.Phase == core.NodeExecution_ABORTED || e.Phase == core.NodeExecution_SKIPPED || e.Phase == core.NodeExecution_TIMED_OUT
+	return i.idempotentRecord(ctx, id, e, e.Phase.String(), terminal)
+}
+
+func (i *idempotentWorkFlowEventRecorder) RecordTaskEvent(ctx context.Context, e *event.TaskExecutionEvent) error {
+	id := e.TaskId.String()
```
EngHabu, line 68: This TaskId doesn't contain the WF execution id I don't think. So this won't be unique within the current execution nor among other executions...

clients/go/events/idempotent_recorder.go
```diff
@@ -0,0 +1,102 @@
...
+	terminal := e.Phase == core.TaskExecution_SUCCEEDED || e.Phase == core.TaskExecution_ABORTED || e.Phase == core.TaskExecution_FAILED
+	return i.idempotentRecord(ctx, id, e, e.Phase.String(), terminal)
+}
+
+// cache of size 5000, ~300KB in size
+func constructIdempotentEventRecorder(eventSink EventSink, scope promutils.Scope) (*idempotentWorkFlowEventRecorder, error) {
+	cfg := GetConfig(context.Background())
```
EngHabu, line 75: nit: pass config as a param to this func... and as a param to the NewIdemponentWorkflo* funcs too.. makes unit testing slightly easier :)

clients/go/events/idempotent_recorder.go
```diff
@@ -0,0 +1,102 @@
...
+
+		i.cache.Add(id, &EmittedEventInfo{lastUpdatedAt: emitTime, phase: phase})
+	}
+
+	return nil
+}
+
+func (i *idempotentWorkFlowEventRecorder) RecordWorkflowEvent(ctx context.Context, e *event.WorkflowExecutionEvent) error {
+	id := e.ExecutionId.String()
+	terminal := e.Phase == core.WorkflowExecution_SUCCEEDED || e.Phase == core.WorkflowExecution_FAILED ||
```
EngHabu, line 55: Do we use terminal in anything?

surindersinghp: Not anymore :). removing. thanks!

clients/go/events/idempotent_recorder.go
surindersinghp: fixed

surindersinghp:
Gentle ping!! @kumare3 @EngHabu @anandswaminathan

clients/go/events/idempotent_recorder.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (i *idempotentWorkFlowEventRecorder) idempotentRecord(ctx context.Context, id string, e proto.Message, phase string) error {
+	var existing *EmittedEventInfo
+	o, found := i.cache.Get(id)
+	if found {
+		existing = o.(*EmittedEventInfo)
+	}
+
+	emitTime := time.Now()
+	if !found || existing.phase != phase || emitTime.Sub(existing.lastUpdatedAt) > time.Duration(i.maxUpdateLagSeconds)*time.Second {
```
EngHabu, line 38: Should we attempt to make it a bit more involved here? e.g. for Array tasks, the task might move to running before we have logs for its subtasks, when they start being constructed, with this change, we will only see the update once every 5 mins...
The alternative might be to also cache the phase version... because that we do change every time we change something...

Should we make the check based on the hashed value of the PhaseInfo we are trying to send or something of that sort?

surindersinghp:
@EngHabu I had addressed your comment after our conversation but forgot to inform you. Can you please take a look?

surindersinghp:
@anandswaminathan fyi

PR #64, opened 2020-05-14
Flyte CoPilot now available for all containers

kumare3:
# TL;DR
Flyte CoPilot is available for all containers instead of a plugin.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

## Complete description
Refer to the issue and associated design

## Tracking Issue
https://github.com/lyft/flyte/issues/297

## Follow-up issue
NA


protos/flyteidl/core/tasks.proto
```diff
@@ -154,4 +154,44 @@ message Container {
     // Ports to open in the container. This feature is not supported by all execution engines. (e.g. supported on K8s but
     // not supported on AWS Batch)
     repeated ContainerPort ports = 7;
+
+    // BETA: This enables use of CoPilot. This makes it possible to to run a completely portable container, that uses inputs and outputs
```
EngHabu, line 158: nit: generally speaking, I do not like using code names in code.... flyte (the product name/identity) is the exception... I would replace CoPilot (also casing looks weird :D) with what it actually does...

protos/flyteidl/core/tasks.proto
```diff
@@ -154,4 +154,44 @@ message Container {
...
+    // only from the local file-system and without having any reference to flyteidl. This is supported only on K8s at the moment.
+    // If CoPilot is enabled, then data will be mounted in accompanying directories specified in the CoPilot settings. If the directories 
+    // are not specified, inputs will be mounted onto and outputs will be uploaded from a pre-determined file-system path. Refer to the documentation
+    // to understand the default paths.
+    bool use_copilot = 8;
```
EngHabu, line 163: Can we put this inside the CoPilot message? and rename it to `enabled`?

protos/flyteidl/core/tasks.proto
```diff
@@ -154,4 +154,44 @@ message Container {
...
+
+    // Optional configuration for CoPilot. If not specified, then default values are used.
+    CoPilot copilot_config = 9;
+}
+
+// This configuration allows executing raw containers in Flyte using the Flyte CoPilot system.
+// Flyte CoPilot, eliminates the needs of flytekit or sdk inside the container. Any inputs required by the users container are side-loaded in the input_path
+// Any outputs generated by the user container - within output_path are automatically uploaded.
+message CoPilot {
```
EngHabu, line 172: Should we call it `MetadataConfig` or `InputOutputMetadataConfig` (too long)? just something descriptive you don't need to read the docs to understand kind of thing

protos/flyteidl/core/tasks.proto
```diff
@@ -154,4 +154,44 @@ message Container {
...
+  // MetadataFormat decides the encoding format in which the input metadata should be made available to the containers. 
+  // If the user has access to the protocol buffer definitions, it is recommended to use the PROTO format.
+  // JSON and YAML do not need any protobuf definitions to read it
+  // All remote references in core.LiteralMap are replaced with local filesystem references (the data is downloaded to local filesystem)
+  enum MetadataFormat {
```
EngHabu, line 177: If you go with `MetadataConfig`, you can shorten this to `Format`

protos/flyteidl/core/tasks.proto
```diff
@@ -154,4 +154,44 @@ message Container {
...
+    // JSON / YAML are serialized represnetation of a map[string]primitive. primitive -> int, string, bool, double/float, bytes
+    JSON = 0;
```
EngHabu, line 179: Can you add a sample JSON structure to the docs here...

protos/flyteidl/core/tasks.proto
```diff
@@ -154,4 +154,44 @@ message Container {
...
+    PROTO = 2;
+  }
+  // File system path (start at root). This folder will contain all the inputs exploded to a separate file. 
+  // Example, if the input interface needs (x: int, y: blob, z: multipart_blob) and the input path is "/var/flyte/inputs", then the file system will look like
+  // /var/flyte/inputs/inputs.<metadata format dependent -> .pb .json .yaml> -> Format as defined previously. The Blob and Multipart blob will reference local filesystem instead of remote locations 
+  // /var/flyte/inputs/x -> X is a file that contains the value of x (integer) in string format
+  // /var/flyte/inputs/y -> Y is a file in Binary format
+  // /var/flyte/inputs/z/... -> Note Z itself is a directory
+  // More information about the protocol - refer to docs #TODO reference docs here
+  string input_path = 1;
```
EngHabu, line 191: should it say `input_dir` ?

setup.py
```diff
@@ -1,6 +1,6 @@
 from setuptools import setup, find_packages
 
-__version__ = '0.17.30'
+__version__ = '0.18.0'
```
EngHabu, line 3: It's not a breaking change... let's keep it at 17.

```suggestion
__version__ = '0.17.31'
```

protos/flyteidl/core/tasks.proto
kumare3: data?

protos/flyteidl/core/tasks.proto
kumare3: isnt that weird? This is like an existence check isnt it?

protos/flyteidl/core/tasks.proto
kumare3: DataConfig?

protos/flyteidl/core/tasks.proto
kumare3: true, will do

setup.py
kumare3: but, if you want to use data loading, you need a min version. That is why i was thinking of upping the minor?

protos/flyteidl/core/tasks.proto
kumare3: as discussed, k8s convention seems to be path https://kubernetes.io/docs/concepts/storage/volumes/#aws-ebs-example-configuration

protos/flyteidl/core/tasks.proto
kumare3: renamed to DataLoadingConfig

protos/flyteidl/core/tasks.proto
kumare3: as the name does not match, we will have to keep metadata format :(

setup.py
EngHabu: hmmm if you want to use data loading, you must be on `0.17.31+`, same thing, right? we should really move to v1 :)

kumare3: ya soon :)

protos/flyteidl/core/tasks.proto
```diff
@@ -153,5 +153,47 @@ message Container {
...
+    // to understand the default paths.
+    // Only K8s
+    DataLoadingConfig data_config = 9;
+}
+
+// This configuration allows executing raw containers in Flyte using the Flyte CoPilot system.
+// Flyte CoPilot, eliminates the needs of flytekit or sdk inside the container. Any inputs required by the users container are side-loaded in the input_path
+// Any outputs generated by the user container - within output_path are automatically uploaded.
+message DataLoadingConfig {
+  // MetadataFormat decides the encoding format in which the input metadata should be made available to the containers.
```
wild-endeavor, line 173: Sorry to be nitpicky, but I've read these comments three times now and I'm still confused.  What is `MetadataFormat`?  The comment says "... format in which the input metadata...". What is input metadata and how is it different than input?  It sounds like it's the same thing.  If it's an input, can we not just call it "Input"?  Currently the situation makes it sound like: 1) there's input data; 2) there's metadata about that input; 3) the format of that metadata can be configured; but 4) the format of the input itself cannot be configured.

Sorry to be annoying but this is the IDL - names here stick around forever.

kumare3: i think we just need to write better documentation about metadata. as this is not really data in all cases.

PR #66, opened 2020-05-29
Adding proto definitions for supporting SageMaker TrainingJob (built-in algorithms) and HyperparameterTuningJob

bnsblue:
# TL;DR
This PR adds the essential proto messages for the support of SageMaker TrainingJob and HyperparameterTuningJob.

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [x] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
This PR adds the essential proto messages for the support of SageMaker TrainingJob and HyperparameterTuning.
The major messages added include `HyperparameterTuningJob`, `TrainingJob`, and `ParameterRanges`. 

The additions allow users to:
1. configure, create, and invoke a TrainingJob
2. configure, create, and invoke a HyperparameterTuningJob
3. specify type-safe hyperparameter ranges for HyperparameterTuningJobs from Flytekit

## Tracking Issue
https://github.com/lyft/flyte/issues/255

## Follow-up issue
https://github.com/lyft/flyte/issues/431


protos/flyteidl/plugins/sagemaker.proto
```diff
@@ -0,0 +1,102 @@
+syntax = "proto3";
+
+package flyte.plugins.sagemaker;
+
+option go_package = "github.com/kumare3/awsflyteplugins/gen/pb-go/proto";
```
EngHabu, line 5: Hmmm I think you should follow the regular path here... of other plugins...

`option go_package = "github.com/lyft/flyteidl/gen/pb-go/flyteidl/plugins";`

protos/flyteidl/plugins/sagemaker.proto
```diff
@@ -0,0 +1,102 @@
...
+    HPOJobObjectiveType Type = 1;
+    string MetricName = 2;
+}
+
+message HPOJobConfig {
+    string Strategy = 1;
+    HPOJobObjective Objective = 2;
+    int64 MaxNumberOfTrainingJobs = 3;
+    int64 MaxParallelTrainingJobs = 4;
+    ParameterRanges ParameterRanges = 5;
```
EngHabu, line 93: Is this really a "config"? not an input of the job?

protos/flyteidl/plugins/sagemaker.proto
```diff
@@ -0,0 +1,102 @@
...
+}
+
+message SagemakerHPOJob {
```
EngHabu, line 96: I presume you meant for this to go into the custom field of the task?

protos/flyteidl/plugins/sagemaker.proto
EngHabu: Since this has a lot of messages, can you actually add one extra level of nesting? 
`option go_package = "github.com/lyft/flyteidl/gen/pb-go/flyteidl/plugins/sagemaker";`

If this works, can you also remove the SageMaker prefix of the messages below?

protos/flyteidl/plugins/sagemaker.proto
bnsblue: This was trying to follow the CRD definition of the amazon-sagemaker-operator-for-k8s. But I think what you suggested here makes more sense.

protos/flyteidl/plugins/sagemaker.proto
bnsblue: Yes

protos/flyteidl/plugins/sagemaker.proto
bnsblue: sure let me give it a try

protos/flyteidl/plugins/sagemaker/sagemaker.proto
```diff
@@ -0,0 +1,98 @@
+syntax = "proto3";
+
+package flyte.plugins.sagemaker;
```
EngHabu, line 3: This should be:
```suggestion
package flyteidl.plugins.sagemaker;
```

bnsblue: good catch! fixed

protos/flyteidl/plugins/sagemaker/trainingjob.proto
```diff
@@ -0,0 +1,45 @@
...
+    int64 volume_size_in_gb = 3;
+    string volume_kms_key_id = 4;
+}
+
+message StoppingCondition {
+    int64 max_runtime_in_seconds = 1;
+    int64 max_wait_time_in_seconds = 2;
+}
+
+message VpcConfig {
```
kumare3, line 32: I think we should leave this out. This should be a backend configuration IMO

protos/flyteidl/plugins/sagemaker/trainingjob.proto
```diff
@@ -0,0 +1,45 @@
...
+    repeated string security_group_ids = 1;
+    repeated string subnets = 2;
+}
+
+message TrainingJob {
+    string region = 1;
+    string role_arn = 2;
```
kumare3, line 39: I think this should be the same role that we are using for the workflow?

protos/flyteidl/plugins/sagemaker/trainingjob.proto
```diff
@@ -0,0 +1,45 @@
...
+    int64 max_wait_time_in_seconds = 2;
+}
+
+message VpcConfig {
+    repeated string security_group_ids = 1;
+    repeated string subnets = 2;
+}
+
+message TrainingJob {
+    string region = 1;
```
kumare3, line 38: I think region should be implicit (and configured in the plugin)

protos/flyteidl/plugins/sagemaker/trainingjob.proto
```diff
@@ -0,0 +1,45 @@
...
+    string algorithm_name = 3;   // TODO: will this change output? should it be an input instead of a part of the task spec?
+
+    message MetricDefinition {
+        string name = 1;
+        string regex = 2;
+    }
+    repeated MetricDefinition metric_definitions = 4;
+}
+
+message ResourceConfig {
```
kumare3, line 20: this is interesting. I want to see how can we remove this from the users perspective too

protos/flyteidl/plugins/sagemaker/trainingjob.proto
```diff
@@ -0,0 +1,45 @@
+syntax = "proto3";
+
+package flyteidl.plugins.sagemaker;
+
+option go_package = "github.com/lyft/flyteidl/gen/pb-go/flyteidl/plugins";
+
+
+message AlgorithmSpecification {
+    string training_image = 1;  // TODO: will this change output? should it be an input instead of a part of the task spec?
```
kumare3, line 9: should we not rely on the image that is part of the task?

protos/flyteidl/plugins/sagemaker/trainingjob.proto
```diff
@@ -0,0 +1,45 @@
...
+    string training_input_mode = 2;
+    string algorithm_name = 3;   // TODO: will this change output? should it be an input instead of a part of the task spec?
```
kumare3, line 11: is this like the HPO algorithm? or the training algorithm. If its training algorithm it will change the output right?

protos/flyteidl/plugins/sagemaker/trainingjob.proto
```diff
@@ -0,0 +1,45 @@
...
+}
+
+message TrainingJob {
+    string region = 1;
+    string role_arn = 2;
+    AlgorithmSpecification algorithm_specification = 3;
+    ResourceConfig resource_config = 4;
+    StoppingCondition stopping_condition = 5;
+    VpcConfig vpc_config = 6;
+    bool enable_spot_training = 7;
```
kumare3, line 44: should we just use interruptible for this?

protos/flyteidl/plugins/sagemaker/trainingjob.proto
bnsblue: This is the training algorithm yes, and changing It will change the output.

protos/flyteidl/plugins/sagemaker/trainingjob.proto
bnsblue: Sure this should follow the convention

protos/flyteidl/plugins/sagemaker/trainingjob.proto
bnsblue: But I am not entirely sure how I should design the user interface (in SDK) for this.. Let me check the pytorch task

protos/flyteidl/plugins/sagemaker/trainingjob.proto
bnsblue: Right. That means we do not need to define this as a part of the message?

protos/flyteidl/plugins/sagemaker/trainingjob.proto
bnsblue: We might have to maintain a lookup table that maps from `(cpu, memory, gpu)x(request, limit)` to `instance type`. But `instance count` seems to be pretty implicit.

protos/flyteidl/plugins/sagemaker/hyperparameter_tuning_job.proto
```diff
@@ -0,0 +1,56 @@
+syntax = "proto3";
+
+package flyteidl.plugins.sagemaker;
+
+option go_package = "github.com/lyft/flyteidl/gen/pb-go/flyteidl/plugins";
+
+import "flyteidl/plugins/sagemaker/parameter_ranges.proto";
+import "flyteidl/plugins/sagemaker/training_job.proto";
+
+// The hyperparameter tuning job
```
EngHabu, line 10: You might want to add references to AWS Reference Docs about the parameters throughout the doc... and explain that these are merely passthrough (or case they are not, explain how we are using them)...

protos/flyteidl/plugins/sagemaker/hyperparameter_tuning_job.proto
```diff
@@ -0,0 +1,56 @@
...
+    // The maximum number of training jobs that an hpo job can launch. For resource limit purpose.
+    int64 max_number_of_training_jobs = 2;
+
+    // The maximum number of concurrent training job that an hpo job can launch
+    int64 max_parallel_training_jobs = 3;
+}
+
+// The objective of the hyperparameter tuning
+message HyperparameterTuningObjective {
+    enum HyperparameterTuningObjectiveType {
```
EngHabu, line 24: I think this should be just called `Type` instead.. otherwise you will reference it by the very long name: `HyperparameterTuningObjective.HyperparameterTuningObjectiveType.MINIMIZE` ... etc.

EngHabu: plz double check my claim in the generated python or by looking at other enums we have already added :)

protos/flyteidl/plugins/sagemaker/hyperparameter_tuning_job.proto
```diff
@@ -0,0 +1,56 @@
...
+    // When the training jobs launched by the hyperparameter tuning job are not improving significantly,
+    // a hyperparameter tuning job can be stopping early.
+    // Note that there's only a subset of built-in algorithms that supports early stopping.
+    // see: https://docs.aws.amazon.com/sagemaker/latest/dg/automatic-model-tuning-early-stopping.html
+    enum TrainingJobEarlyStoppingType {
+        OFF = 0;
+        AUTO = 1;
+    }
+    TrainingJobEarlyStoppingType training_job_early_stopping_type = 4;
+}
```
EngHabu, line 56: nit:
```suggestion
}

```

protos/flyteidl/plugins/sagemaker/parameter_ranges.proto
```diff
@@ -0,0 +1,49 @@
+syntax = "proto3";
+
+package flyteidl.plugins.sagemaker;
+
+option go_package = "github.com/lyft/flyteidl/gen/pb-go/flyteidl/plugins";
```
EngHabu, line 5: Does this still correctly generate a `"github.com/lyft/flyteidl/gen/pb-go/flyteidl/plugins/sagemaker"` package following the file path? or will these end up in `"github.com/lyft/flyteidl/gen/pb-go/flyteidl/plugins"` package ?

protos/flyteidl/plugins/sagemaker/parameter_ranges.proto
```diff
@@ -0,0 +1,49 @@
...
+
+// HyperparameterScalingType defines the way to increase or decrease the value of the hyperparameter
+// https://docs.aws.amazon.com/sagemaker/latest/dg/automatic-model-tuning-define-ranges.html
+enum HyperparameterScalingType {
```
EngHabu, line 9: You probably want to wrap this with a message... otherwise I think the enum name (`HyperparameterScalingType`) will be dropped in some of the generated languages...

```suggestion
message HyperparameterScalingType {
   enum Values {
   
```

EngHabu: but again, double check the other enums we have :-)

protos/flyteidl/plugins/sagemaker/parameter_ranges.proto
```diff
@@ -0,0 +1,49 @@
...
+        ContinuousParameterRange continuous_parameter_range = 1;
+        IntegerParameterRange integer_parameter_range = 2;
+        CategoricalParameterRange categorical_parameter_range = 3;
+    }
+}
+
+// ParameterRanges is a map that maps hyperparameter name to the corresponding hyperparameter range
+message ParameterRanges {
+    map<string, ParameterRangeOneOf> parameter_range_map = 1;
+}
```
EngHabu, line 49: nit:
```suggestion
}

```

protos/flyteidl/plugins/sagemaker/training_job.proto
```diff
@@ -0,0 +1,70 @@
+syntax = "proto3";
+
+package flyteidl.plugins.sagemaker;
+
+option go_package = "github.com/lyft/flyteidl/gen/pb-go/flyteidl/plugins";
+
+
+enum InputMode {
```
EngHabu, line 8: ditto

protos/flyteidl/plugins/sagemaker/training_job.proto
```diff
@@ -0,0 +1,70 @@
...
+    // training job runs, so it must be equal to or greater than max_runtime_in_seconds.
+    int64 max_wait_time_in_seconds = 2;
+}
+
+// The spec of a training job
+message TrainingJob {
+    AlgorithmSpecification algorithm_specification = 1;
+    TrainingJobConfig training_job_config = 2;
+    bool interruptible = 3;
+}
```
EngHabu, line 70: nit
```suggestion
}

```

protos/flyteidl/plugins/sagemaker/hyperparameter_tuning_job.proto
```diff
@@ -0,0 +1,56 @@
...
+message HyperparameterTuningSpecification {
+
+    ParameterRanges hyperparameter_ranges = 1;
+    enum HyperparameterTuningStrategy {
+        BAYESIAN = 0;
+        RANDOM = 1;
+    }
+
+    HyperparameterTuningStrategy tuning_strategy = 2;
+    HyperparameterTuningObjective tuning_objective = 3;
```
EngHabu, line 45: so you can specify multiple metrics in the training job algorithm spec but you can only have one metric used for HPO? is that correct?

protos/flyteidl/plugins/sagemaker/hyperparameter_tuning_job.proto
bnsblue: You are right it is really long and that bothered me lol

protos/flyteidl/plugins/sagemaker/parameter_ranges.proto
bnsblue: The package is generated at the right location `"github.com/lyft/flyteidl/gen/pb-go/flyteidl/plugins/sagemaker"`

protos/flyteidl/plugins/sagemaker/hyperparameter_tuning_job.proto
bnsblue: Yes, according to SM official document

protos/flyteidl/plugins/sagemaker/parameter_ranges.proto
bnsblue: To summarize the issue: if the enum is not wrapped inside a message, the generated python code for an enum, say,
```proto
enum FixedRateUnit {
    MINUTE = 0;
    HOUR = 1;
    DAY = 2;
}
```
will be orphans exposed at the module level
```python
_schedule_pb2.MINUTE
```
https://github.com/lyft/flytekit/blob/master/flytekit/models/schedule.py#L10

When we have multiple of them in a proto file, you can see that the generated python file can be confusing, especially when there are name collisions such as 
```
enum A {
   AUTO = 0;
   OFF = 1;
}

enum B {
  LINEAR = 0;
  AUTO = 1;
}
```
So the way to resolve this issue is to wrap every enum inside a message

protos/flyteidl/plugins/sagemaker/training_job.proto
```diff
@@ -0,0 +1,84 @@
...
+
+    // The maximum length of time in second that the training job can run.
+    // If this value is not specified, the default expiration time will be 1 day
+    int64 max_runtime_in_seconds = 1;
+
+    // The maximum length of time in seconds that the users are willing to wait for a managed spot
+    // training job to complete.
+    // Note that it is the amount of time spent waiting for Spot capacity plus the amount of time the
+    // training job runs, so it must be equal to or greater than max_runtime_in_seconds.
+    int64 max_wait_time_in_seconds = 2;
```
EngHabu, line 75: Should we use proto Duration type instead?

protos/flyteidl/plugins/sagemaker/hyperparameter_tuning_job.proto
```diff
@@ -0,0 +1,71 @@
...
+
+option go_package = "github.com/lyft/flyteidl/gen/pb-go/flyteidl/plugins";
+
+import "flyteidl/plugins/sagemaker/parameter_ranges.proto";
+import "flyteidl/plugins/sagemaker/training_job.proto";
+
+// A pass-through for SageMaker's hyperparameter tuning job
+message HyperparameterTuningJob {
+    // The underlying training job that the hyperparameter tuning job will launch during the process
+    TrainingJob training_job = 1;  // TODO: This hierarchical definition is tricky!
```
kumare3, line 13: TODO and the description do not go well together. reword?

protos/flyteidl/plugins/sagemaker/training_job.proto
```diff
@@ -0,0 +1,76 @@
+syntax = "proto3";
+
+package flyteidl.plugins.sagemaker;
+
+import "google/protobuf/duration.proto";
+
+option go_package = "github.com/lyft/flyteidl/gen/pb-go/flyteidl/plugins";
+
+
+message InputMode {
```
kumare3, line 10: some description?

generate_protos.sh
```diff
@@ -2,7 +2,7 @@
 
 DIR=`pwd`
 rm -rf $DIR/gen
-LYFT_IMAGE="lyft/protocgenerator:5e6a3be18db77a8862365a19711428c2f66284ef"
+LYFT_IMAGE="lyft/protocgenerator:8167e11d3b3439373c2f033080a4b550078884a2"
```
kumare3, line 5: is this intentional?

bnsblue: Yes. See this: https://github.com/lyft/protocgenerator/commit/8167e11d3b3439373c2f033080a4b550078884a2

kumare3:
Lets just mostly remove the `TODO` or `word it better` we are then good to go!

PR #67, opened 2020-06-02
Add FailureHandlingStrategy to workflow idl

EngHabu:
# TL;DR
Adds a failure handling strategy flag to workflow Metadata. This defines how the system should behave when it encounters a failing node in the workflow. 

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added
 - [X] Any pending items have an associated Issue

## Tracking Issue
https://github.com/lyft/flyte/issues/191


protos/flyteidl/core/workflow.proto
```diff
@@ -127,11 +126,30 @@ message Node {
...
+
+        // WAIT_FOR_EXECUTABLE_BRANCHES instructs the system to make as much progress as it can. The system will not
+        // alter the dependencies of the execution graph so any node that depend on the failed node will not be run.
+        // Other nodes that will be executed to completion before cleaning up resources and marking the workflow
+        // execution as failed.
+        WAIT_FOR_EXECUTABLE_BRANCHES = 2;
+    }
+
+    // Defines how the system should behave when a failure is detected in the workflow execution.
+    FailureHandlingStrategy failure_handling_strategy = 2;
```
kumare3, line 152: will this be overridable in the launch-plan?

protos/flyteidl/core/workflow.proto
```diff
@@ -127,11 +126,30 @@ message Node {
...
 
-// Metadata for the entire workflow.
-// To be used in the future.
+// Metadata for the entire workflow. Defines execution behavior that does not change the final outputs of the workflow.
 message WorkflowMetadata {
     // Total wait time a workflow can be delayed by queueing.
     google.protobuf.Duration queuing_budget = 1;
+
+    // Failure Handling Strategy
+    enum FailureHandlingStrategy {
```
kumare3, line 135: how about OnFailure
FAIL_IMMEDIATELY
FAIL_AFTER_
FAIL_ON_RUNNING_NODES_DONE

protos/flyteidl/core/workflow.proto
EngHabu: done

protos/flyteidl/core/workflow.proto
EngHabu: better?

protos/flyteidl/admin/launch_plan.proto
```diff
@@ -52,63 +53,73 @@ message LaunchPlanList {
...
+        string assumable_iam_role = 1;
+        string kubernetes_service_account = 2;
+    }
 }
 
+// Defines allowed overrides for the workflow template.
+message WorkflowTemplateOverrides {
+    // If a field is set in metadata, it'll override the one from the original workflow template.
+    // If it's not-set, the system will continue to use the value from the original workflow template.
+    // TODO: Should we just require a full replacement of metadata to avoid having to deal with merges?
```
schottra, line 71: This seems like something that should be decided before moving ahead with the changes. It might be hard to switch from merge to wholesale replacement once records exist using the partial data format.
Thoughts: 
* Merge is easier on the user
* Wholesale replacement records would be rendered incomplete when we add new fields to the metadata. I know protobuf has an 'all things are optional' approach, but practically speaking that's not always true. We would need to consider what it means for a new metadata field to be added and any existing launch plans to, by convention, exclude the value because they are replacing the entire metadata object and don't specify the value.

EngHabu: @kumare3 @wild-endeavor thoughts?

wild-endeavor: With respect to the specifica question at hand, I'd prefer to do a merge.  It's just slightly trickier, it's not exactly rocket science, and can be very fully unit tested.  Not worried about this at all.

What is the goal of this by the way?  Is this the start of a general solution for execution overrides and will one day include information on task level cpu/memory overrides, concurrency overrides, etc.?  What this PR has makes sense for the run to completion flag, but I haven't thought through how task level overrides will work in the future, and whether this approach will mesh with those needs seamlessly.

wild-endeavor:
Also bump setup.py

protos/flyteidl/admin/launch_plan.proto
kumare3: This is an interesting question.  So @bnsblue just brought up and some users have, about the ability to override node metadata. This PR allows node metadata.
I know I added this as a comment, but lets hold off on this change. I think this deserves a small spec and I think @bnsblue has started recording some thoughts.

WDYT?

kumare3: also my two cents, merging is not a simple endeavor, replacement has the con that Randy mentioned, of overriding fields. But if the older spec does not consist it, newer one should override any non existent fields too right?

bnsblue: I am linking the one-pager doc for your reference. I just started it yesterday so it is still pretty brief at the moment: https://docs.google.com/document/d/1bA9T_DHSE2_752N5Y-K8AtO4VZQ69aJDzGn274SkC6I/edit#

PR #68, opened 2020-06-08
Add quality of service to launch plan/execution spec

katrogan:
# TL;DR
This PR introduces a QualityOfService designation for workflow executions. This is primarily used to designate toleration for queueing time but can potentially be leveraged for other server-side execution configuration.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
https://docs.google.com/document/d/1rYqr01x0XL6n1AdmEOUbTIGeS9Z2hJCcnG5NMy8aQ68/edit#

## Tracking Issue
https://github.com/lyft/flyte/issues/343

## Follow-up issue
_NA_


migueltol22:
Now that we are adopting quality of service should we keep https://github.com/lyft/flyteidl/blob/42929ee56bc48fc887ed372b468929d797f39643/protos/flyteidl/core/workflow.proto#L132

katrogan:
if queueing_budget is unused, happy to remove it!

protos/flyteidl/admin/common.proto
```diff
@@ -245,3 +245,18 @@ message AuthRole {
   }
 }
 
+// This general categorization specifies how much queue time an execution can tolerate. 
+// This can be extended in the future to make other execution-time decisions.
+enum QualityOfService {
```
EngHabu, line 250: I think we adopted this model instead:
```suggestion
message QualityOfService {
    enum Values {
        UNDEFINED = 0;
        ...
    }
}
```

protos/flyteidl/admin/execution.proto
```diff
@@ -237,6 +237,8 @@ message ExecutionSpec {
     // Optional: auth override to apply this execution.
     AuthRole auth_role        = 16;
 
+    // Indicates the amount of queueing a launched workflow execution can tolerate.
```
EngHabu, line 240: Is it only about queuing? does it also control interruptible bit?

protos/flyteidl/admin/common.proto
```diff
@@ -245,3 +245,18 @@ message AuthRole {
...
+    // Default: no quality of service specified.
+    QUALITY_OF_SERVICE_UNDEFINED = 0;
+
+    // Guarantees that this execution will begin as soon as requested and incur no queueing time.
+    QUALITY_OF_SERVICE_HIGH = 1;
+
+    // This execution may incur some queueing delay (e.g. 30 minutes) and is medium priority.
```
anandswaminathan, line 257: Do not indicate time here in comments. The range currently is huge.

katrogan:
@EngHabu thanks for the suggestions. I started a doc [here](https://docs.google.com/document/d/1rYqr01x0XL6n1AdmEOUbTIGeS9Z2hJCcnG5NMy8aQ68/edit?usp=sharing) to capture a few different implementation proposals and open questions.

katrogan:
@EngHabu @migueltol22 @anandswaminathan mind taking another look? updated to reflect design decisions/discussions

katrogan:
ping @EngHabu @migueltol22 @anandswaminathan

protos/flyteidl/admin/quality_of_service.proto
```diff
@@ -0,0 +1,28 @@
+syntax = "proto3";
+
+package flyteidl.admin;
+option go_package = "github.com/lyft/flyteidl/gen/pb-go/flyteidl/admin";
+
+
+message QualityOfServiceSpec {
+  // Indicates how much queueing delay an execution can tolerate.
+  uint32 queueing_budget_mins = 1;
```
EngHabu, line 9: Should we use [this](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Duration) instead?

protos/flyteidl/admin/launch_plan.proto
```diff
@@ -91,6 +92,9 @@ message LaunchPlanSpec {
     Auth auth                             = 8 [deprecated=true];
 
     AuthRole auth_role                    = 9;
+
+    // Indicates the amount of queueing a launched workflow execution can tolerate.
+    QualityOfService quality_of_service   = 16;
```
migueltol22, line 97: Since this will be part of the launchplan should we also remove [this](https://github.com/lyft/flyteidl/blob/e279a7eb92b56f726fe13795e4f088976b14f752/protos/flyteidl/core/workflow.proto#L132)?

katrogan: I believe we'll want to resolve the queueing budget based on the quality of service on the admin side before creating the workflow CRD so we should still keep the queueing_budget field

protos/flyteidl/admin/quality_of_service.proto
katrogan: originally that's what I had but people commented that anything beyond minutes was too much granularity and unnecessary

one option is we could expose it as queueing_budget_mins in the SDK and then store that as a duration in the proto field but that can still be manipulated by non-SDK users

protos/flyteidl/admin/launch_plan.proto
katrogan: thank you for the catch! updated to use QualityOfService in the workflow metadata

protos/flyteidl/admin/quality_of_service.proto
katrogan: will go with the second option, updated to use duration and we can expose it as a field in mins only in the sdk cc @migueltol22

PR #70, opened 2020-06-11
feature; catalog metadata and caching status published in node events

kumare3:
# TL;DR
Caching Information is currently not available in the UI or CLI. This hinders user productivity. This PR sets up the flow of data between propeller and admin. Once the data is published, next step would be to expose it through the API

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Tracking Issue
https://github.com/lyft/flyte/issues/138

## Follow-up issue
https://github.com/lyft/flyte/issues/71


kumare3:
> any versions you need to bump?

Updated now

protos/flyteidl/core/identifier.proto
```diff
@@ -10,6 +10,7 @@ enum ResourceType {
   TASK               = 1;
   WORKFLOW           = 2;
   LAUNCH_PLAN        = 3;
+  DATASET            = 4;
```
wild-endeavor, line 13: One of these things is not like the other.  Can we at least add a comment explaining why there is something that looks different here?  Also, could we create a GH issue if there isn't one to merge all the proto objects in data catalog https://github.com/lyft/datacatalog/blob/master/protos/idl/service.proto into this repo?

kumare3: Agreed

PR #71, opened 2020-07-02
tensorflow.proto and respective changes

pingsutw:
# TL;DR
proto files and generated ones for TensorFlow Flyte plugin

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [x] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
N/A

## Follow-up issue
N/A


protos/flyteidl/plugins/tensorflow.proto
```diff
@@ -0,0 +1,13 @@
...
+
+package flyteidl.plugins;
+
+option go_package = "github.com/lyft/flyteidl/gen/pb-go/flyteidl/plugins";
+
+// Custom proto for plugin that enables distributed training using https://github.com/kubeflow/tf-operator
+message DistributedTensorflowTrainingTask {
+  // number of worker, ps, chief replicas spawned in the cluster for this job
+  int32 workers = 1;
+  int32 ps = 2;
```
kumare3, line 11: Can we put a command `PS -> Parameter server?`

kumare3: also should it be "ps_replicas"

protos/flyteidl/plugins/tensorflow.proto
```diff
@@ -0,0 +1,13 @@
...
+  int32 chief = 3;
```
kumare3, line 12: Also a comment to explain what "chief" means. Should it be "chief_replicas". I acutally do not know this, can "chief" have different "min" and "max". From the CRD it seems that it can only be `1`. In that case can we just omit this value here
https://github.com/kubeflow/tf-operator/blob/master/examples/crd/crd-v1.yaml#L37

In FlytePlugin it is possible to provide this as a configuration. Here is an example - https://github.com/lyft/flyteplugins/blob/master/go/tasks/plugins/k8s/spark/spark.go#L37

This value is configured as here
https://github.com/lyft/flyte/blob/master/kustomize/overlays/sandbox/propeller/plugins/spark/config.yaml#L2

Reason being, we can make it simpler for the user. But, ofcourse this depends on whether "chief" can have different values

kumare3:
@pingsutw Thank you for the PR, just a couple comments. Otherwise this looks good. Ping me and we can merge it soon

protos/flyteidl/plugins/tensorflow.proto
pingsutw: The chief replicas number could be 0. 
TF has a `multi-worker` distributed strategy, they will only launch worker nodes without ps and chief.
example: https://github.com/kubeflow/tf-operator/blob/master/examples/v1/distribution_strategy/keras-API/multi_worker_tfjob.yaml
TF-operator will check spec whether contains **Chief** 
https://github.com/kubeflow/tf-operator/blob/24798375fead22bb1a78f3039565ccf5a9fae017/pkg/controller.v1/tensorflow/status.go#L89-L142

pingsutw:
@kumare3 Thanks for the review.
I just updated my PR.
update include:
- rename protobuf 
ps -> ps_replicas, 
chief -> chief_replicas
- add a comment for ps

pingsutw:
> Lgtm, qq; do the workers etc need mem/cpu?

I think it should be added, but I found `mem/cpu` also don't include in PyTorch proto.
In order to make it consistency, I didn't add them. Maybe we could add it later if I run into some problems in other PRs.   
https://github.com/lyft/flyteidl/blob/master/protos/flyteidl/plugins/pytorch.proto

katrogan:
hey @pingsutw mind incrementing the version [here](https://github.com/lyft/flyteidl/blob/master/package.json#L3) and [here](https://github.com/lyft/flyteidl/blob/master/setup.py#L3) as part of this change?

pingsutw:
@katrogan Thanks for the reminder.
The new version is `0.17.35`?

pingsutw:
I've updated the version.

PR #72, opened 2020-07-21
IDL changes for Node-Node relationship refactor

anandswaminathan:
This PR introduces the following
- Ability to provide information of the parent node.
- Introduction of group_id to indicate grouping of nodes within the Parent node - (instead of dummy nodes)
- Support for querying all nodes under node with group.
- Support for node_id to point to original node in the workflow/graph, and node_name

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Tracking Issue
https://github.com/lyft/flyte/issues/218


anandswaminathan:
PTAL @kumare3 @katrogan @schottra

protos/flyteidl/admin/node_execution.proto
```diff
@@ -35,6 +35,10 @@ message NodeExecutionListRequest {
     // Sort ordering.
     // +optional
     Sort sort_by    = 5;
+
+    // Unique identifier of the parent node in the execution
+    // +optional
+    string unique_parent_id = 6;
```
katrogan, line 41: why is this a string and not NodeExecutionIdentifier?

protos/flyteidl/event/event.proto
```diff
@@ -62,8 +62,20 @@ message NodeExecutionEvent {
...
     ParentTaskExecutionMetadata parent_task_metadata         = 9;
+
+    // Specifies the parent node of the current node execution. Node executions at level zero will not have a parent node.
+    ParentNodeExecutionMetadata parent_node_metadata	= 10;
+
+    // Group identifier to indicate arbitrary grouping like retries, sub workflow
+    string group_id = 11;
+
+    // Identifier of the node in the original workflow/graph
+    string node_id = 12;
```
katrogan, line 75: isn't this originally captured by the node_id in the identifier?

protos/flyteidl/event/event.proto
```diff
@@ -62,8 +62,20 @@ message NodeExecutionEvent {
...
 
-    // Specifies which task (if any) launched this node.
+    // [To be deprecated] Specifies which task (if any) launched this node.
     ParentTaskExecutionMetadata parent_task_metadata         = 9;
+
+    // Specifies the parent node of the current node execution. Node executions at level zero will not have a parent node.
+    ParentNodeExecutionMetadata parent_node_metadata	= 10;
+
+    // Group identifier to indicate arbitrary grouping like retries, sub workflow
+    string group_id = 11;
```
katrogan, line 72: question: should this be part of the node execution identifier?

protos/flyteidl/event/event.proto
```diff
@@ -75,6 +87,11 @@ message ParentTaskExecutionMetadata {
     core.TaskExecutionIdentifier id = 1;
 }
 
+message ParentNodeExecutionMetadata {
+    // Unique identifier of the parent node id within the execution
+    string id = 1;
```
katrogan, line 92: this references the `node_id`?

do we anticipate this changing over time ever? would it make sense to make it a proper proto message with just one field for now?

protos/flyteidl/admin/node_execution.proto
anandswaminathan: Nope. I did not want that. Why duplicating everything, when we know that nodes cannot be across execution?

protos/flyteidl/event/event.proto
anandswaminathan: Unique identifier of the node is the NodeExecutionId - this is just a label/id to group node executions.

protos/flyteidl/event/event.proto
anandswaminathan: Yes.  That will now become unique identifier. This will point to the original node id in the graph.

protos/flyteidl/event/event.proto
anandswaminathan: This is the unique_node_id (we are moving from node_id to unique node id). I am open to suggestions. Merely following the `ParentTaskExecutionMetadata` model. But fine with anything.

protos/flyteidl/event/event.proto
katrogan: is this label unique within the execution?

protos/flyteidl/event/event.proto
schottra: I didn't see this field exposed in any of the NodeExecution structures above. Is that intentional? As it is, I don't see a way for consumers of NodeExecution to know which node in the graph it relates to.

protos/flyteidl/admin/node_execution.proto
schottra: I don't see the endpoint definition in this changeset, so I'm not sure how this new field will be used when querying?
I guess I'm wondering why we are choosing to have it as a separate field rather than a filter?

protos/flyteidl/admin/node_execution.proto
```diff
@@ -79,6 +83,12 @@ message NodeExecution {
 
 // Represents additional attributes related to a Node Execution
 message NodeExecutionMetaData {
+    // Node executions are grouped depending on retries of the parent, and other scenarios
+    // Group id is unique within the context of a parent node.
+    string group_id = 1;
```
schottra, line 88: I take this to imply that if I query for the children of a NodeExecution, I will get all children across all attempts and then I can sort them into buckets by this value on the client side?

protos/flyteidl/event/event.proto
anandswaminathan: @katrogan It's not - as in need not be. It only need to be unique within a parent node. But that's up to implementation, and future use cases.

protos/flyteidl/admin/node_execution.proto
anandswaminathan: Exactly. @schottra  This is to prevent dummy nodes, and help with the visualization. We will return all executions and you can use the group id to group and show.

protos/flyteidl/admin/node_execution.proto
anandswaminathan: True. Removing it. This will be just another filter.

protos/flyteidl/event/event.proto
kumare3: so effectively, there would be no enforcement or semantic checking on this field?

kumare3: but If I want to query all node-executions for a parent node, then I could get the parent node and find the group id?

protos/flyteidl/event/event.proto
kumare3: what is the `unique_node_id` -> Is it not the nodeexecutionIdentifier?

anandswaminathan: @kumare3 It is the id in `nodeexecutionIdentifier `.

nodeexecutionIdentifier has node_id. nodeexecutionIdentifier also has WorkflowExecutionIdentifier- which project,domain, name

protos/flyteidl/event/event.proto
anandswaminathan: @kumare3 What do you mean semantic checking? 

If you want all executions for a parent node, you will simply pass the filter with the value of unique_node_id in it. Group id is for retries inside the parent node.

protos/flyteidl/event/event.proto
kumare3: then should this not be an object / structure?

anandswaminathan: That's what I mentioned to @katrogan. I do not want that. It will only cause unwanted project, domain, name checks. 

You can't have a parent node from a different execution.

protos/flyteidl/event/event.proto
```diff
@@ -62,8 +62,21 @@ message NodeExecutionEvent {
...
     }
 
-    // Specifies which task (if any) launched this node.
+    // [To be deprecated] Specifies which task (if any) launched this node.
     ParentTaskExecutionMetadata parent_task_metadata         = 9;
+
+    // Specifies the parent node of the current node execution. Node executions at level zero will not have a parent node.
+    ParentNodeExecutionMetadata parent_node_metadata	= 10;
+
+    // Retry group to indicate arbitrary grouping like retries, sub workflow
```
kumare3, line 71:
```suggestion
    // Retry group to indicate grouping of nodes by retries
```

protos/flyteidl/admin/node_execution.proto
schottra: We discussed this offline and decided a separate endoint/request parameter is more appropriate than a filter, since specifying this value will return the base set of results and not filter down the set of results that would be returned if you left the value empty.

PR #73, opened 2020-07-22
Return full execution inputs & outputs

katrogan:
# TL;DR
This change allows users to specify they want full inputs & outputs in a Get{Workflow,Node,Task}ExecutionData request but preserves existing functionality by defaulting to only return signed url blobs.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
This is useful to return inputs & outputs in sandbox mode when the signed url is host-specific

## Tracking Issue
https://github.com/lyft/flyte/issues/419

## Follow-up issue
_NA_


katrogan:
ping @kumare3 @wild-endeavor

protos/flyteidl/admin/task_execution.proto
```diff
@@ -101,6 +102,9 @@ message TaskExecutionClosure {
 message TaskExecutionGetDataRequest {
     // The identifier of the task execution for which to fetch inputs and outputs.
     core.TaskExecutionIdentifier id        = 1;
+
+    // Specifies what type of data to return in the response.
+    GetExecutionData.Mode mode             = 2;
```
kumare3, line 107: what is the default mode? Should we make the default as returning results inline?

protos/flyteidl/admin/common.proto
```diff
@@ -245,3 +245,14 @@ message AuthRole {
   }
 }
 
+// Specifies how to populate Get{Workflow,Node,Task}ExecutionData responses.
+message GetExecutionData {
```
kumare3, line 249: so the reason why I suggested we make this change is because - 

1. Metadata is entirely owned by Flyte. Returning a signed URL is actually a leaky abstraction and forces us to leak the detail that the metadata is stored in S3. A future world, we may store the metadata in DynamoDB for example.
2. One problem with always returning the data is that, sometimes the data is very large and that has problems in FlyteAdmin.

WDYT? if we always return the entire metadata (as literal map) and also the signed url (for backwards compatibility). If the data is larger than some threshold (can be verified using HEAD request), then we only return the signed URL.
Then we start tracking how many times this happens and start figuring out a solution to eliminate that case

protos/flyteidl/admin/task_execution.proto
katrogan: I made the default mode url blobs only to keep with existing behavior.

protos/flyteidl/admin/common.proto
katrogan: sure, we're already issuing the HEAD request to capture the data size for the URLBlob portion of the response anyways. I can get rid of the mode and return the data always when it's under a configured limit.

schottra: I would love if we could get to a solution where we aren't signing URLs if the user doesn't absolutely need to use them. I think from Console's perspective, it would be preferable to see either the data or the URLs. And I would be okay if that's something I have to opt in to with a flag of some sort.
But if that feels like it's over-complicated, or there's no real concern about issuing a signed URL every time we request execution data, please disregard!

katrogan: I don't think including a signed URL adds too much overhead (we have to do the HEAD request either way) so I'm okay returning it by default but if we want to phase it out, happy to make this configurable by a flag

@kumare3 thoughts?

kumare3: so @schottra and @katrogan I think eventually all data that we return should fit within the limits. As we should start offloading lists and long strings should just be outright rejected. So in this scenario, it would be that we always return data inline. SO I would love to move in that direction, this makes it work perfectly in all conditions and hides the complexity in Admin

katrogan: cool, so for the sake of backwards compatibility while we migrate to that model, what if we just always return the signed URL and the full data?

katrogan:
mind taking another look @kumare3 @wild-endeavor @schottra ?

protos/flyteidl/admin/task_execution.proto
```diff
@@ -110,4 +111,10 @@ message TaskExecutionGetDataResponse {
 
     // Signed url to fetch a core.LiteralMap of task execution outputs.
     UrlBlob outputs                            = 2;
+
+    // Optional, full_inputs will only be populated if they are small enough to be returned
```
kumare3, line 115: can we define small enough?

katrogan: sure, done

PR #78, opened 2020-08-12
Introduce ProjectUpdate endpoint to Flyte Admin API

kosigz-lyft:
# TL;DR
Introduce ProjectUpdate endpoint to Flyte Admin API.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added


kosigz2:
PTAL #flyte-eng @katrogan

protos/flyteidl/admin/project.proto
```diff
@@ -44,3 +44,10 @@ message ProjectRegisterRequest {
 
 message ProjectRegisterResponse {
 }
+
+message ProjectUpdateRequest {
+    Project project                       = 1;
```
katrogan, line 49: Can you add comments on how this updates/patches existing projects? if you're only trying to add a label, does your Project need to include the existing description for that field not to get overwritten to empty?

protos/flyteidl/service/admin.proto
```diff
@@ -313,6 +313,16 @@ service AdminService {
     };
   }
 
+  rpc UpdateProject (flyteidl.admin.ProjectUpdateRequest) returns (flyteidl.admin.ProjectUpdateResponse) {
+    option (google.api.http) = {
+      post: "/api/v1/projects"
```
schottra, line 318: If it's an update, can we use `put` or `patch` instead of `post`? `post`ing to this endpoint should have the result of creating a new project.

protos/flyteidl/admin/project.proto
schottra: REST semantics would say that if we use the endpoint as defined below (`api/v1/projects`) with a `PUT,` then we pass the entire object and replace the whole thing all at once. If we intend to merge values into the record, then we would use a `PATCH` and the minimum required properties on the object would be an `id` plus at least one other field to be merged into the target record. `PUT` of an entire record is a little cumbersome because it often requires a separate `GET` operation first if you don't have the entire project. It's also slightly more prone to errors from race conditions or stale data.

But also worth thinking about, if we allow wholesale or partial replacement of the entire record, that means we have to support adding/removing domains as part of this request. That has other implications, right?

We may want to do this similar to how we update metadata for `NamedEntities` (workflows/tasks/etc)
https://github.com/lyft/flyteidl/blob/master/protos/flyteidl/admin/common.proto#L142

Following that scheme, the URL would be something like `api/v1/projects/{id}` or if the endpoint is just for updating metadata (is updating the list of domains something we want to allow?), then something like `api/v1/project_metadata/{id}`.

Also, with the entity metadata updates, the shape of the request object is `id` + whatever fields you want to allow an update for. And passing any of those values has the effect of replacing that field completely in the target record. We could consider doing the same thing here.

PR #80, opened 2020-09-14
Create PluginOverrides MatchableResource.

katrogan:
# TL;DR
This change introduces a mechanism for configuring a custom plugin implementation for a task type that can be set at the domain, project+domain, or project, domain + workflow level.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
https://docs.google.com/document/d/1OWYJ2ggr0pWvOdOKkf04NLtj2Rf32dMVYvAYMoyeU48/edit?usp=sharing

## Tracking Issue
https://github.com/lyft/flyte/issues/499

## Follow-up issue
_NA_


protos/flyteidl/admin/matchable_resource.proto
```diff
@@ -58,6 +61,29 @@ message ExecutionClusterLabel {
   string value	= 1;
 }
 
+message PluginOverride {
```
kumare3, line 64: lets have dome documentation on the message structure itself?

kumare3:
This looks awesome @katrogan, thank you. One comment

protos/flyteidl/admin/matchable_resource.proto
katrogan: good point, done

protos/flyteidl/admin/matchable_resource.proto
```diff
@@ -58,6 +61,32 @@ message ExecutionClusterLabel {
...
+
+    // Use the in-order list of fallback_plugin_ids as supported fallbacks for this task type.
+    OVERRIDE_FALLBACK = 1;
+
+    // Uses the system-configured default implementation.
+    USE_DEFAULT = 2;
+  }
+  MissingPluginBehavior missing_plugin_behavior = 4;
+}
+
```
EngHabu, line 88:
```suggestion
  // A set of plugin ids which should handle tasks of this type instead of the default registered plugin. The list will be tried in order until a plugin is found with that id.
  repeated string plugin_id = 2;

  enum MissingPluginBehavior {
    FAIL = 0;

    // Uses the system-configured default implementation.
    USE_DEFAULT = 1;
  }
  
  // Defines the behavior when no plugin from the plugin_id list is not found.
  MissingPluginBehavior missing_plugin_behavior = 4;
}

```

EngHabu: I'm ok with either option.. can you think of any reason to go with one option over the other?

katrogan: nope, your suggestion is certainly less verbose! thanks

katrogan:
friendly ping @EngHabu @kumare3

kumare3:
(courtesy @wild-endeavor ) One potential problem (I say potential, because it may not be a problem at all), is that when Propeller launches a launch plan, new plugin defaults could be used, in case of subworkflow same would be used. So potential of different behavior.

But, the solution for this is the story we have in the backlog. Task events should have plugin_id, task_type and other information sent back to admin

katrogan:
@kumare3 good point, but this is similarly an issue we already have to contend with with say inconsistently applied task resource defaults, execution queues and other matchable resources. we should probably determine what we want the expected behavior to be across the board

protos/flyteidl/admin/matchable_resource.proto
```diff
@@ -58,6 +61,28 @@ message ExecutionClusterLabel {
...
+    // Uses the system-configured default implementation.
+    USE_DEFAULT = 1;
+  }
+  
+  // Defines the behavior when no plugin from the plugin_id list is not found.
+  MissingPluginBehavior missing_plugin_behavior = 4;
+}
+
+
```
EngHabu, line 85:
```suggestion
```

EngHabu (approved):

katrogan:
one last approval please?

EngHabu (approved):

PR #82, opened 2020-09-23
Add option to record whether a task ran on a spot instance.

katrogan:
# TL;DR
Add option to record whether a task ran on a spot instance.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
https://docs.google.com/document/d/1nqAp5-DanklNvG6OeQ3uIacZYGvf0oXioz7CboP_CsU/edit

## Tracking Issue
https://github.com/lyft/flyte/issues/520

## Follow-up issue
_NA_


protos/flyteidl/event/event.proto
```diff
@@ -150,4 +150,20 @@ message TaskExecutionEvent {
...
+message TaskExecutionMetadata {
+    // Includes the type of machine used for this specific task execution. 
+    enum InstanceType {
+        // The default instance type configured for the flyte application platform.
+        DEFAULT = 0; 
+
+        // The instance type configured for interruptible tasks.
+        INTERRUPTIBLE = 1;
+    }
+    InstanceType instance_type = 1;
```
wild-endeavor, line 168: can it ever run on both?  if it gets interrupted three time and the fourth time we use a non-spot?

katrogan: yep it can, but that's why this is per task execution. if we retry on a non-spot node that retry attempt will be a distinct task execution and the instance type will change for its metadata

migueltol22 (approved):
LGTM

protos/flyteidl/event/event.proto
```diff
@@ -150,4 +150,20 @@ message TaskExecutionEvent {
     // that should be recorded regardless of the lack of phase change.
     // The version field should be incremented when metadata changes across the duration of an individual phase.
     uint32 phase_version                  = 12;    
+
+    // Metadata around how a task was executed.
+    TaskExecutionMetadata metadata        = 16;
+}
+
+// Holds metadata around how a task was executed.
+message TaskExecutionMetadata {
```
kumare3, line 159: This is great, can we conform to - https://github.com/lyft/flyte/issues/325

protos/flyteidl/event/event.proto
```diff
@@ -150,4 +150,20 @@ message TaskExecutionEvent {
...
+    // Includes the type of machine used for this specific task execution. 
+    enum InstanceType {
```
kumare3, line 161: Why is this an enum? Is it because interruptible is a top level field in Task. I think that is fine, 
I would like to add other things here like - TaskType, TaskPluginID
and then a map<string,string> of other data

protos/flyteidl/event/event.proto
katrogan: hm I can add a freeform map as well? is there a specific reason this shouldn't have any structure?

protos/flyteidl/event/event.proto
katrogan: because i specifically wanted to define the set of possible values for instance type in particular :)

katrogan:
@kumare3 I'm not sure I understand why we want the task execution metadata to be specifically a free form map? in any case, if using an enum for instance type is acceptable can I keep this PR as is and let whoever takes on https://github.com/lyft/flyte/issues/325 handle updating the proto message accordingly?

protos/flyteidl/event/event.proto
EngHabu, line 168: Is it called "Instance Type" does that not refer to the `p316xlarge` stuff?

EngHabu (dismissed):

protos/flyteidl/event/event.proto
katrogan: perhaps InstanceClass would be more appropriate (and less specific?) if it's preferable to use the p316xlarge stuff, then the freeform map may make more sense here too

kumare3 (approved):

PR #83, opened 2020-09-29
Support Styx style schedule

honnix:
# TL;DR
When integrating with our scheduler Styx, we will need to extend the semantics of schedule to support Styx cron syntax, cron aliases, as well as offset.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Complete description and details can be found in the [issue description](https://github.com/lyft/flyte/issues/529).

## Tracking Issue
https://github.com/lyft/flyte/issues/529


katrogan:
hey @honnix mind bumping the version [here](https://github.com/lyft/flyteidl/blob/master/setup.py#L3) and [here](https://github.com/lyft/flyteidl/blob/master/package.json#L3) as part of your change?

honnix:
Sure I will fix that.

protos/flyteidl/admin/schedule.proto
```diff
@@ -16,6 +16,13 @@ message FixedRate {
     FixedRateUnit unit = 2;
 }
 
+// Styx style schedule
+// More details can be found at https://github.com/spotify/styx#schedule-string
+message StyxSchedule {
```
kumare3, line 21: Should we call it StyxSchedule or just `CronScheduleWithOffset`
We should also add a note here
```python
Supported only by some schedulers for example Spotify/StyxSchedule. Your platform may not have this enabled.
And correspondingly we should fail registration if using a schedule that is not supported
```

honnix: Totally. This would look a lot better. I will fix that.

honnix:
I made the changes as suggested. I will work on flyteadmin later to reject launch plan with this kind of schedule, then probably there is no need to change flyteconsole to display the schedule. Another PR will be made in flytekit to propagate the proto.

honnix:
I forgot to `make generate` again. :( Sorry and will fix.

honnix:
Does this look more reasonable now?

protos/flyteidl/admin/schedule.proto
```diff
@@ -16,6 +16,12 @@ message FixedRate {
     FixedRateUnit unit = 2;
 }
 
+// Supported only by some schedulers for example Spotify/Styx. Your platform may not have this enabled.
+message CronScheduleWithOffset {
+    string schedule = 1;
```
kumare3, line 21: is the schedule here actually the expression?
Also, this is pretty good, i think eventually we should just deprecate cron_expression for CronScheduleWithOffset (just offset will be 0 by default and not supported on most platforms)

honnix: It can be a cron expression or alias such as hourly, daily, weekly, monthly etc. So we were not so sure what name could be more reasonable.

honnix: I think this PR https://github.com/lyft/flytekit/pull/188 gives more details.

protos/flyteidl/admin/schedule.proto
```diff
@@ -16,6 +16,12 @@ message FixedRate {
     FixedRateUnit unit = 2;
 }
 
+// Supported only by some schedulers for example Spotify/Styx. Your platform may not have this enabled.
+message CronScheduleWithOffset {
```
wild-endeavor, line 20:
```suggestion
message CronSchedule {
```

protos/flyteidl/admin/schedule.proto
```diff
@@ -24,6 +30,7 @@ message Schedule {
         // e.g. for a schedule that runs every 15 minutes: "0/15 * * * ? *"
         string cron_expression = 1;
         FixedRate rate = 2;
+        CronScheduleWithOffset cron_schedule_with_offset = 4;
```
wild-endeavor, line 33:
```suggestion
        CronSchedule cron_schedule = 4;
```

protos/flyteidl/admin/schedule.proto
```diff
@@ -24,6 +30,7 @@ message Schedule {
         // e.g. for a schedule that runs every 15 minutes: "0/15 * * * ? *"
         string cron_expression = 1;
```
wild-endeavor, line 31:
```suggestion
        string cron_expression = 1 [deprecated=true];
```

wild-endeavor (commented):
not sure what @katrogan thinks?

honnix:
@wild-endeavor I guess it is legacy reason, the cron expression handling is very weird in flytekit as it basically casts from a richer semantics to a weaker one by string manipulation. Is it OK with this change we start accepting Unix flavoured cron in launch plan definition? For example, for the deprecated cron expression we keep the the existing logic unchanged, while for CronSchedule we parse it directly using croniter, and the validation will also allow cron alias, such as daily, weekly, etc. The validation logic can be found in the proposed flytekit PR.

honnix:
@wild-endeavor What do you think of the name of this class: https://github.com/lyft/flytekit/blob/6274318ad9f64b0367754a917d08d0f133358300/flytekit/common/schedules.py#L128. There is already a class named `CronSchedule`.

honnix:
I have a second thought now, that I think it might make other changes easier (flytekit and flyteadmin) to name the message as `CronScheduleWithOffset`, without conflicting with existing names that we need to keep for backward compatibility.

wild-endeavor:
+1 after changing the name to `CronSchedule`

katrogan:
I like CronSchedule too! I think it will make sense to move to that as the replacement for cron_expression (which like @honnix mentioned is a weird/less-structured format anyways)

honnix:
How do we handle the python class name then? We could have more validation branches to cover alias, Unix cron and existing cron in the same class. How does that sound?

honnix:
@katrogan @wild-endeavor I revert it back to `CronSchedule`. Please take a look again. Thanks.

protos/flyteidl/admin/schedule.proto
```diff
@@ -16,14 +16,20 @@ message FixedRate {
     FixedRateUnit unit = 2;
 }
 
+message CronSchedule {
+    string schedule = 1;
+    string offset = 2;
```
katrogan, line 21: one nit: can you add comments on what the expected format for schedule and offset should be?

katrogan (approved):
LGTM, @wild-endeavor mind taking one last look?

protos/flyteidl/admin/schedule.proto
honnix: @katrogan Do we plan to support AWS cron as well in `CronSchedule.schedule`? If yes, I can revise the comment a bit.

wild-endeavor (approved):

protos/flyteidl/admin/schedule.proto
katrogan: I think we should move away from the aws cron type (which is more work) and have admin handle converting to it in its aws specific scheduler handler. But that can be a separate issue and we can keep using the deprecated cron_expression in the meantime

honnix: I see. The problem is the 5-fields cron syntax is weaker than the one used by AWS, so it's not gonna be an 100% mapping.

honnix:
Alright. I'm merging this and will make changes in flytekit accordingly. Thanks everyone for reviewing and commenting on this PR.

honnix:
@katrogan I drafted this release https://github.com/lyft/flyteidl/releases/tag/untagged-eaee7bcfa18de386fe2b, please take a look and release if it is OK.

katrogan:
@honnix looks good!

honnix:
Somehow the github workflow is not triggered to publish to pypi. :thinking:

wild-endeavor:
I just tried to push it manually - can you try to pull on your end?

honnix:
@wild-endeavor Yeah that worked. I guess I messed it by drafting a release and then publish. There are weird things regarding which event to use: https://stackoverflow.com/questions/59319281/github-action-different-between-release-created-and-published

PR #88, opened 2020-11-04
Added state to project

honnix:
# TL;DR
Added state attribute to project protobuf

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Added `state` attribute to project so when updating a project it can be archived or active.

## Tracking Issue
https://github.com/lyft/flyte/issues/569

## Follow-up issue
_NA_


protos/flyteidl/admin/project.proto
```diff
@@ -29,6 +29,8 @@ message Project {
     // Leverage Labels from flyteidel.admin.common.proto to
     // tag projects with ownership information.
     Labels labels = 5;
+
+    bool archived = 6;
```
katrogan, line 33: what do you think on adding a state enum like https://github.com/lyft/flyteidl/blob/c3baba8983019680ef57b6244cea36ba951233ed/protos/flyteidl/admin/common.proto#L25? not sure if overkill for entities like project but even for sandbox deploys could be nice to one day have a system-generated type attribute too

honnix: Yeah makes sense. 👍 Didn't know this existed. Thanks for pointing out.

honnix: Does it make sense to use that enum directly than creating a new one? What is the semantics or usage of that enum?

honnix: Hmm probably not according to the comment of NamedEntity.

katrogan: yeah, maybe better to use a new enum definition in this case.

protos/flyteidl/admin/project.proto
```diff
@@ -15,6 +15,18 @@ message Domain {
     string name  = 2;
 }
 
+// The status of the project is used to control its visibility in the UI and validity.
+enum ProjectState {
```
katrogan, line 19: to avoid the fun of the repeated usage of the word 'project' in the generated enum values, can we wrap the enum in a message like https://john-millikin.com/effective-grpc#enums (I know NamedEntity does not follow this convention, unfortunately)

katrogan (commented):
looks good! mind bumping the version in setup.py and package.json?

honnix:
@katrogan sure. Do we follow semantic versioning strictly? If yes this case it's gonna be `0.19.0` then. Or should we still stay on `0.18.x`.

katrogan (approved):

PR #90, opened 2020-11-24
Add filter and sort param when listing projects

honnix:
# TL;DR
Add filter and sort param when listing projects.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Currently when listing projects, it is not supported to provide filters and sort param, also it is not possible to do pagination. This PR largely copies `ResourceListRequest` to support those filters.

## Tracking Issue
https://github.com/lyft/flyte/issues/625

## Follow-up issue
_NA_


katrogan:
looks good! mind bumping the version(s)?

katrogan (approved):

honnix:
@katrogan Not exactly sure why CI failed. I did commit all changes.

katrogan:
@EngHabu it looks like https://github.com/lyft/flyteidl/commit/9ffba7eb6c44324cc44248fb2ec332d890ff7dcb breaks tests - can you look into this?

EngHabu:
I'll take a look in 30 mins if that's ok

honnix:
OK, I see. Sure, please take the time. Thanks for looking into this.

EngHabu:
Plz review: https://github.com/lyft/flyteidl/pull/91

katrogan (approved):

katrogan:
cc @schottra I think the console will now need to add a limit and deal with pagination in list project requests

schottra:
@katrogan Thanks for the ping, we definitely would have to add pagination support to console.

@honnix I'd like to discuss the use case for this. Flyteconsole is currently designed with the idea that the list of projects will be retrieved *once*, in full. The two places where we list the projects do not have any affordance for pagination and those experiences would have to be addressed.

honnix:
@schottra This new endpoint is designed to be backward compatible, so if no offset and limit provided, it should return all projects.

honnix:
A bit more context on this feature. As a left over of https://github.com/lyft/flyte/issues/569, in the beginning I only planed to add filter and sorting support, then I realised it is really little effort to add pagination as well. The API is designed in a way to be backward compatible with existing usage (where no query parameter would be provided). Whether to support pagination on console is completely a user experience decision, because everything would still work as it is.

I believe it is fair to imagine in some user cases the number of projects could be huge, so it can potentially degrade the experience quite a bit, both console and CLI.

honnix:
Can I merge this and continue with https://github.com/lyft/flyteadmin/pull/142 ? @schottra @katrogan 

Thanks.

schottra:
@honnix 
I appreciate making it backwards-compatible so that the console does not have to update in order to continue functioning.

> Whether to support pagination on console is completely a user experience decision, because everything would still work as it is.
This is what I was getting at. I'm trying to decide what the user experience should be. If you imagine your company's usage of Flyte creating so many projects that pagination is necessary (150+?), then the console experience will degrade if we don't update it. If you are just adding it to make API interactions more flexible and don't expect to have more than a few dozen projects, then I am okay leaving it alone for now.

PR #102, opened 2021-02-23
Add task version to template

katrogan:
# TL;DR
Adds task version to task template

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/768

## Follow-up issue
_NA_


protos/flyteidl/core/tasks.proto
```diff
@@ -118,6 +118,9 @@ message TaskTemplate {
     oneof target {
         Container container = 6;
     }
+
+    // This can be used to customize task handling at execution time for the same task type.
+    int32 version = 7;
```
EngHabu, line 123: Should we name it explicitly `task_type_version` ?
Should this be a string? to use SemVer?

katrogan: re: Should we name it explicitly task_type_version - sure, wasn't sure if it would be a mouthful given that it's in the TaskTemplate but probably better to be explicit.

not sure if semver is necessarily required here. what do you think?

protos/flyteidl/core/tasks.proto
```diff
@@ -118,6 +118,9 @@ message TaskTemplate {
     oneof target {
         Container container = 6;
     }
+
+    // This can be used to customize task handling at execution time for the same task type.
+    int32 task_type_version = 7;
```
kumare3, line 123: can we no call it version? or protocol_version?

EngHabu: I don't think we should overload what it means... Version is confusing with task id version.. this is more about the spec...

protos/flyteidl/core/tasks.proto
EngHabu: Will it get to a place where we want to define compatibility ranges? Like v2 is backwards compatible with v1 but v3 is a breaking change... 

Or does this sound too complicated?

protos/flyteidl/core/tasks.proto
katrogan: @kumare3 I originally had it as version but I'm inclined to agree with Haytham, this is meta info regarding the task type. Happy to change this again but not sure what you both prefer? `task_spec_version`?

protos/flyteidl/core/tasks.proto
katrogan: that sounds reasonable, but is semver semantics necessary to achieve backwards compatibility or support ranges? i'm hesitant to add rich/complex versioning if there's no anticipated use-cases (but I'm not sure if I'm missing any)

EngHabu: It's just an understood concept... right? if you look at a task template and you see the version is "1.0.0" then it changes (with a flytekit update or something) to "1.1.0", you can safely assume it should continue to work with the same backend... if it just jumps from int 2 to int 3, you can't really tell...

but I also have to agree with you, I don't have a scenario in mind for the flexible ranges... 

This is the library we use in stdlib https://github.com/coreos/go-semver for parsing and comparing semversions...

I don't feel strongly either way... your call...

EngHabu (approved):

protos/flyteidl/core/tasks.proto
kumare3: I guess we have a field called as `type` in this proto, so lets keep it `task_type_version`

protos/flyteidl/core/tasks.proto
katrogan: right, it is definitely convention but for things like event version we've also just gone with a simple int. given that historically we've made limited breaking-type changes I'm inclined to keep it simple for now and if it ends up being restrictive maybe we can revisit?

cool to know there's a library to make parsing semver easy though!

PR #119, opened 2021-03-16
#patch: Add in task execution event fields

katrogan:

# TL;DR
Add in task execution event fields

- Add **phase reason** and **task type** to TaskExecutionEvent
- Add **generated_name** to TaskExecutionMetadata
- Add a **ManagedResourceInfo** message for capturing resource token name (and future FRM attributes)

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/325

## Follow-up issue
_NA_


protos/flyteidl/event/event.proto
```diff
@@ -167,4 +185,10 @@ message TaskExecutionMetadata {
         INTERRUPTIBLE = 1;
     }
     InstanceClass instance_class = 16;
+
+    // Generated unique name for this task execution used by the backend.
+    string generated_name = 1;
+
+    // Includes information about how resource token allocation (if applicable).
+    ManagedResourceInfo managed_resource_info = 2;
```
EngHabu, line 193: should this be repeated?

protos/flyteidl/event/event.proto
```diff
@@ -151,12 +151,30 @@ message TaskExecutionEvent {
...
+// This message holds task execution metadata specific to resource allocation used to manage concurrent
+// executions for a project namespace.
+message ManagedResourceInfo {
+    // Unique resource ID used to identify this execution when allocating a token.
+    string token                         = 1;
+}
+
 // Holds metadata around how a task was executed.
 // TODO(katrogan): Extend to include freeform fields (https://github.com/flyteorg/flyte/issues/325).
+// Next ID to use:  3
```
EngHabu, line 177: While I liked these back in the day, unless we have tools to make sure they are correct (do we?) they will get out of sync and be confusing... I think.

protos/flyteidl/event/event.proto
```diff
@@ -151,12 +151,30 @@ message TaskExecutionEvent {
...
     uint32 phase_version                  = 12;    
 
+    // If there is an explanation for this phase transition, the reason will capture it.
+    string reason                         = 13;
+
+
+    // A predefined yet extensible Task type identifier. If the task definition is already registered in flyte admin
+    // this type will be identical, but not all task executions necessarily use pre-registered definitions and this
+    // type is useful to render the task in the UI, filter task executions, etc.
+    string type                           = 14;
```
EngHabu, line 161: Is this the task type? or the event type? from the comment it seems it refers to the task type... maybe it should be renamed `task_type` ?

katrogan: ah good point, done

protos/flyteidl/event/event.proto
katrogan: fair point, done

protos/flyteidl/event/event.proto
katrogan: perhaps i'm misunderstanding but isn't this only determined once per execution?

EngHabu: A plugin can request multiple resource allocations during execution... our plugins only request one resource at most because we didn't need more... but the API allows plugins to request more

katrogan: @EngHabu got it, thanks. updated

protos/flyteidl/event/event.proto
```diff
@@ -151,10 +151,27 @@ message TaskExecutionEvent {
...
     // Metadata around how a task was executed.
     TaskExecutionMetadata metadata        = 16;
 }
 
+
+// This message holds task execution metadata specific to resource allocation used to manage concurrent
+// executions for a project namespace.
+message ManagedResourceInfo {
+    // Unique resource ID used to identify this execution when allocating a token.
+    string token                         = 1;
```
EngHabu, line 172: Should we call this `allocation_token` ?

EngHabu: Should we capture more of the request? https://github.com/flyteorg/flyteplugins/blob/master/go/tasks/pluginmachinery/core/resource_manager.go#L82
the resource namespace maybe?

EngHabu (dismissed):

protos/flyteidl/event/event.proto
katrogan: sure done @EngHabu

EngHabu (dismissed):

protos/flyteidl/event/event.proto
```diff
@@ -167,4 +187,11 @@ message TaskExecutionMetadata {
         INTERRUPTIBLE = 1;
     }
     InstanceClass instance_class = 16;
+
+    // Generated unique name for this task execution used by the backend.
+    string generated_name = 1;
```
kumare3, line 192: should we call it resource_name?
As for Qubole/AWS Batch etc, it may not be the generated_name?
Or we can have both - but that is confusing!

katrogan: done

protos/flyteidl/event/event.proto
```diff
@@ -151,10 +151,40 @@ message TaskExecutionEvent {
     // The version field should be incremented when metadata changes across the duration of an individual phase.
     uint32 phase_version                  = 12;    
 
+    // If there is an explanation for this phase transition, the reason will capture it.
```
kumare3, line 154:
```suggestion
    // An optional explanation for the phase transition
```

protos/flyteidl/event/event.proto
```diff
@@ -151,10 +151,40 @@ message TaskExecutionEvent {
...
+    string reason                         = 13;
+
+
+    // A predefined yet extensible Task type identifier. If the task definition is already registered in flyte admin
+    // this type will be identical, but not all task executions necessarily use pre-registered definitions and this
+    // type is useful to render the task in the UI, filter task executions, etc.
+    string task_type                      = 14;
```
kumare3, line 161: should we also add the actual execution plugin-ID used?

protos/flyteidl/event/event.proto
```diff
@@ -167,4 +197,14 @@ message TaskExecutionMetadata {
         INTERRUPTIBLE = 1;
     }
     InstanceClass instance_class = 16;
+
+    // Contains various identifiers for resources used during execution.
```
kumare3, line 201: nit: can we order them by the field id? sorry!

katrogan: @kumare3 done

protos/flyteidl/event/event.proto
katrogan: done

katrogan:
friendly ping @EngHabu @kumare3

protos/flyteidl/event/event.proto
```diff
@@ -151,13 +151,54 @@ message TaskExecutionEvent {
...
+    string allocation_token               = 1;
+
+    // Namespace under which this task execution requested an allocation token.
+    string namespace                      = 2;
+}
+
 // Holds metadata around how a task was executed.
 // TODO(katrogan): Extend to include freeform fields (https://github.com/flyteorg/flyte/issues/325).
 message TaskExecutionMetadata {
+
```
kumare3, line 188: Can we remove the todo on the message? indicates task -
https://github.com/flyteorg/flyte/issues/325

protos/flyteidl/event/event.proto
```diff
@@ -151,13 +151,54 @@ message TaskExecutionEvent {
...
 
+    // An optional explanation for the phase transition.
+    string reason                         = 13;
+
+
+    // A predefined yet extensible Task type identifier. If the task definition is already registered in flyte admin
+    // this type will be identical, but not all task executions necessarily use pre-registered definitions and this
+    // type is useful to render the task in the UI, filter task executions, etc.
+    string task_type                      = 14;
+
```
kumare3, line 162: @katrogan another question, why is task_type / reason not metadata? Maybe a clearer definition in the distinction

protos/flyteidl/event/event.proto
```diff
@@ -151,13 +151,54 @@ message TaskExecutionEvent {
...
     // Metadata around how a task was executed.
     TaskExecutionMetadata metadata        = 16;
 }
 
+// This message contains metadata about external resources produced or used by a specific task execution.
+message ExternalResourceInfo {
+
+    // Identifiers for external resources created by this task execution, for example Qubole query ID or presto query ids.
+    repeated string external_ids = 1;
```
EngHabu, line 171: If we have a repeated field [here](https://github.com/flyteorg/flyteidl/pull/119/files#diff-61c99ca77e3f94bf6d64036e11e8851dd5a9defc8fb5f1d385af4bfc52b9e9f4R197) we don't need this to be repeated, right? this can include other things if the plugin wants to (maybe cost related, maybe remote cluster/region info ... etc.) that we can open up later

protos/flyteidl/event/event.proto
katrogan: So I like the documentation at the beginning of this proto message: `Plugin specific execution event information`  i.e. attributes of the event itself. In this case task_type is an attribute of the execution and reason goes along with phase and phase version as part of the actual event update, right? I see metadata is more for identifying/discerning how a task was executed rather than artifacts or products of the event itself.

protos/flyteidl/event/event.proto
katrogan: oh yes good catch, done

protos/flyteidl/event/event.proto
katrogan: done

kumare3 (dismissed):

katrogan:
one last (?) ping @kumare3 @EngHabu

kumare3 (approved):

EngHabu (approved):
👍

PR #132, opened 2021-03-30
Added workflow for updating flyteidl version in all component

evalsocket:
# TL;DR

- Added GitHub action cron for updating flyteidl version in all component

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Tracking Issue
https://github.com/lyft/flyte/issues/868

## Follow-up issue
_NA_
OR


evalsocket:
I added cron rather than release event because we create a release in each merge in master and frequent merge in master will create an overlap of PRs.  It would be great we can run it once/twice a day.

.github/workflows/release-automation.yml
```diff
@@ -0,0 +1,24 @@
...
+    # TODO: Need to discuss
+    - cron:  '5 4 * * *'
+
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        REPOSITORY: [flyteadmin, flytepropeller, flyteconsole, flytekit]
```
EngHabu, line 14: We do need to merge all the other PRs first before this, right?

.github/workflows/release-automation.yml
```diff
@@ -0,0 +1,24 @@
+name: Update flyteidl version 
+
+on:
+  schedule:
+    # TODO: Need to discuss
```
EngHabu, line 5: Shouldn't this happen on release creation?

EngHabu: If you use something like this: https://github.com/marketplace/actions/dependencies-autoupdate it will make sure it doesn't recreate a PR but rather just update the same branch with the latest.... so it'll keep it as a single PR until we merge it then start creating a new one for the next update... etc. would that help?

.github/workflows/release-automation.yml
evalsocket: yes

EngHabu (approved):
🤞 thank you for doing this

PR #142, opened 2021-04-14
Add GetOrReserveArtifact and ExtendReservation API

milton0825:
# TL;DR
Add a `GetOrReserveArtifact` API in datacatalog.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

Currently tasks with same task signature and same input and cache enabled can run in parallel even though they are expect to produce the same output. Ideally we want only one task to proceed with the execution and the rest of the task waiting for the cache result. Thus, we introduce a new `GetOrReserveArtifact` API in datacatalog and tasks are required to call this API to reserve a spot for the execution and this ensures that only one task can run at a time.

The task that holds a reservation should call `ExtendReservation` API periodically to keep the reservation. Otherwise, it can expire and other tasks can take over it.

More details in:
https://docs.google.com/document/d/1Oq8f0yVoHMHDCUbxhu9V-4dDLzbbM4YWcnNNscEnYkM/edit?usp=sharing

## Tracking Issue
https://github.com/flyteorg/flyte/issues/872

## Follow-up issue
_NA_


protos/flyteidl/datacatalog/datacatalog.proto
```diff
@@ -93,6 +95,26 @@ message ListDatasetsResponse {
     string next_token = 2;
 }
 
+message AcquireLeaseRequest {
```
kumare3, line 98: So datacatalog and lease semantics look weird. I think we should alter the existing API to provide optimistic locking - reservations.

The API is invoked from flytepropeller here - https://github.com/flyteorg/flytepropeller/blob/master/pkg/controller/nodes/task/catalog/datacatalog/datacatalog.go#L115

How about we create a new GetArtifact to be
GetOrReserveArtifact ?

milton0825:
@anandswaminathan @kumare3 @akhurana001 ready for review.

protos/flyteidl/datacatalog/datacatalog.proto
```diff
@@ -150,6 +157,22 @@ message ListDatasetsResponse {
...
+    google.protobuf.Timestamp expire_at = 3;
+}
+
+enum ReservationStatus {
+    SUCCESS = 0;
+    FAILED = 1;
+}
+
+message GetOrReserveArtifactResponse {
+    Artifact artifact = 1;
```
kumare3, line 172: this should be a one of?

protos/flyteidl/datacatalog/datacatalog.proto
```diff
@@ -150,6 +157,22 @@ message ListDatasetsResponse {
     string next_token = 2;
 }
 
+message GetOrReserveArtifactRequest {
```
kumare3, line 160: also can you rebase? I add a bunch of documentation to all the structures, you can get that too :)

protos/flyteidl/datacatalog/datacatalog.proto
```diff
@@ -150,6 +157,22 @@ message ListDatasetsResponse {
...
+    DatasetID dataset_id = 1;
+    string tag_name = 2;
+    google.protobuf.Timestamp expire_at = 3;
+}
+
+enum ReservationStatus {
+    SUCCESS = 0;
```
kumare3, line 167: we will have to indicate that a Reservation is already in progress right?

So consider this example

Process 1:
   GetOrReserve(A) -> Reservation(Success)

Process 2:
   GetOrReserve(A) -> Reservation(Failed) - whats the reason for failure?

Also I feel, the timeout should be evaluated on the server side, so how about updating the method

GetOrReserve(A, timeout)

And the Status should be more descriptive
1. ReservationAcquired, ReservationAlreadyInProgress, ReservationFailed, ReservationTimedOut

If the case of ReservationAlreadyInProgress, we should probably also return the holding task.

I would recommend changing
GetOrReserveArtifactRequest to include Metadata - https://github.com/flyteorg/flyteidl/blob/master/protos/flyteidl/datacatalog/datacatalog.proto#L215

Thus the ReservationStatus is not just an enum, but more like
message ReservationStatus {
   enum State {
      ...
   },
   Metadata md
}

milton0825: re: `ReservationTimedOut` 

I think if the reservation expired, the rest of the tasks should just fight for the spot. One task will get `ReservationAcquired` and the rest will get `AlreadyInProgress`

protos/flyteidl/datacatalog/datacatalog.proto
milton0825: Yeah i have merged master

milton0825:
@kumare3 should be ready for another look

protos/flyteidl/datacatalog/datacatalog.proto
```diff
@@ -150,6 +158,38 @@ message ListDatasetsResponse {
     string next_token = 2;
 }
 
+// Get the Artifact or try to reserve a spot if the Artifact does not exist.
+message GetOrReserveArtifactRequest {
+    DatasetID dataset_id = 1;
+    string tag_name = 2;
+
+    // How long you can occupy the spot.
```
kumare3, line 166: Another interesting question over here is, what happens if its been a long time and the process 2 now has been granted a reservation. Lets assume process 3 comes along, will it get an Already_in_progress? If so, we will have to store both the reservations.

I think it is worth documenting these semantics clearly.

Also  this should be reworded
```suggestion
    // Minimum duration to wait from the time when the original reservation was made to consider that the original reservation has been timed-out.
```

kumare3:
@milton0825 thank you for the work, one more comment (mostly around docs) and then we are good to go!

kumare3:
For FlyteIDL, we are not instructing a process of getting atleast 2 +1's 
cc @anandswaminathan, @EngHabu, @kanterov, @honnix either of you have comments on this? Another +1 and we are good to go

protos/flyteidl/datacatalog/datacatalog.proto
```diff
@@ -33,6 +34,13 @@ service DataCatalog {
 
     // Return a paginated list of datasets
     rpc ListDatasets (ListDatasetsRequest) returns (ListDatasetsResponse);
+
+    // Get an artifact and the corresponding data. If the artifact does not exist,
+    // try to reserve a spot for populating the artifact. We may have multiple
+    // concurrent tasks with the same signature and the same input that try to populate
+    // the same artifact at the same time. Thus with reservation, only one task
+    // can run at a time.
```
kumare3, line 42:
```suggestion
    // can run at a time, until the timeout period. Once the timeout has elapsed, another task (Random order) may start executing and a new-timeout duration will be applied from this tasks start time. If the first task completes after the second task has started, a third task may receive the artifact from the first task and the second task may override this artifact. As currently the last writer wins!
```

protos/flyteidl/datacatalog/datacatalog.proto
```diff
@@ -150,6 +158,39 @@ message ListDatasetsResponse {
     string next_token = 2;
 }
 
+// Get the Artifact or try to reserve a spot if the Artifact does not exist.
+message GetOrReserveArtifactRequest {
```
EngHabu, line 162: Should we have sort of an owner_id? so that if the same owner keeps showing up to reserve, it can respond with Acquired... and refresh the timeout

protos/flyteidl/datacatalog/datacatalog.proto
```diff
@@ -150,6 +158,39 @@ message ListDatasetsResponse {
...
+    DatasetID dataset_id = 1;
+    string tag_name = 2;
+
+    // Minimum duration to wait from the time when the original reservation
+    // was made to consider that the original reservation has been timed-out.
+    google.protobuf.Duration timeout = 3;
```
EngHabu, line 168: I don't know what's the plan for using this? How can we predict how long a task will take? I think we can have a config value for this timeout not per request... and expect the owner to keep "heartbeating" using its own unique id before the timeout expires...

protos/flyteidl/datacatalog/datacatalog.proto
milton0825: So the clock starts when a task successfully "reserve" a spot, the task then continue with the execution. If the task does not finish within the timeout, other tasks can take the reservation and kick off another execution.

protos/flyteidl/datacatalog/datacatalog.proto
kumare3: the timeout is specified by the user. It is not a prediction. Heartbeating is usually better in case you want an absolute lease. Here we are saying that this is a best effort model of trying to serialize an execution, but in the worst case, we may have concurrent executions, which is OK, as the only thing this is useful for is cost optimization. 

Heartbeating will need a change in the way propeller talks with catalog and the contract, which seems dangerous?

milton0825: So even we go with the "heartbeating" approach, we still need to specify the task execution timeout because we probably don't want a single healthy but slow task blocking other tasks from making progress. So this come down to what we want to do when encounter a timeout:
1. Kill the task execution and release the lease (task execution timeout)
2. Let the execution continue and release the lease (cache reservation timeout)

I think 1 may be an overkill for the purpose of reducing duplicated effort for populating cache since we don't necessarily need to kill the process.

kumare3: @milton0825 you are right, 1 is not the right approach, because this is not a leader election, this is purely an optimization. Haytham and I had a chat so here is an update

I agree with @EngHabu that asking the users to specify the timeout is error-prone. So instead we use a heartbeat interval like internal configuration. But then, we will need flytepropeller to keep extending the reservation

Thus the pseudo code in propeller

process 1 - gets the reservation, process -2 will get blocked

```python
before task starts:
if caching enabled:
  if cache_serialization_enabled:
        r, err = getCacheOrReservation(md=metadata, provenance-info, reservation_owner:str, 
                  reservation_timeout:duration,)
         if err { then retry with backoff }
         if r.acquired {
              goto start-task
         }
       else return not_acquired("waiting for cache")
    else: // cache serialization not enabled, but caching is
       Getcache(..) //old way

startTask:
     if caching_enabled and caching_serializatin_enabled:
           ExtendReservation(owner_id, provenance, duration)
```

`Important note.`
GetOrReserve API is invoked multple times by the same ownerID with same provenance, then only the timeout is extended.
If the timeout is expired, then GetOrReserve simply returns a new reservation (hence we need an owner id). But the semantics are crisp, there can 2 reservations that have the same provenance still in progress concurrently.
Thus if process one comes back and says, updateReservation, Datacatalog will obey it.

`Open question`
We should probably add a new API that is called Extendreservation(owner-id, timeout)

Also Flytepropeller will need a chance, to always call extendreservation and the timeout should be dynamic in  flytepropeller, this will allow controlling how often this api is invoked

kumare3: Gist:
1. Add reservation_owner and reservation_timeout as the 2 fields
2. >> Add new API ExtendReservation()
3. Document the semantics

milton0825: > GetOrReserve API is invoked multiple times by the same ownerID with same provenance, then only the timeout is extended.

What do ownerId and provenance each refer to in this context? My understanding is that we need an ID to uniquely identify a task execution that "owns" the reservation

protos/flyteidl/datacatalog/datacatalog.proto
```diff
@@ -150,6 +173,47 @@ message ListDatasetsResponse {
     string next_token = 2;
 }
 
+// Get the Artifact or try to reserve a spot if the Artifact does not exist.
+message GetOrReserveArtifactRequest {
+    DatasetID dataset_id = 1;
+    string tag_name = 2;
+    string owner_id = 3;
```
milton0825, line 180: I dropped the timeout from request here since it will be a global timeout and I think we can make it configurable via some config file in datacatalog.

milton0825:
@kumare3 @EngHabu ready for another look

protos/flyteidl/datacatalog/datacatalog.proto
```diff
@@ -33,6 +34,28 @@ service DataCatalog {
...
+    // try to reserve a spot for populating the artifact.
+    //
+    // Once you preserve a spot, you should call ExtendReservation API periodically
+    // to extend the reservation. Otherwise, the reservation can expire and other
+    // tasks may take the spot.
+    //
+    // Note: We may have multiple concurrent tasks with the same signature
+    // and the same input that try to populate the same artifact at the same time.
+    // Thus with reservation, only one task can run at a time, until the reservation
+    // expire.
```
EngHabu, line 48:
```suggestion
    // expires.
```

protos/flyteidl/datacatalog/datacatalog.proto
```diff
@@ -33,6 +34,28 @@ service DataCatalog {
...
+    //
+    // Note: If the task A does not extend the reservation in time and the reservation
+    // may expire. Another task B may take over the reservation and now we have two tasks
+    // A and B running in parallel. So a third task C may get the Artifact from A or B,
+    // whoever is the last writer.
```
EngHabu, line 53:
```suggestion
    // Note: If the task A does not extend the reservation in time and the reservation
    // expires, another task B may take over the reservation and now we have two tasks
    // A and B running in parallel. So a third task C may get the Artifact from A or B,
    // whichever writes last.
```

protos/flyteidl/datacatalog/datacatalog.proto
```diff
@@ -33,6 +34,28 @@ service DataCatalog {
...
+    rpc GetOrReserveArtifact (GetOrReserveArtifactRequest) returns (GetOrReserveArtifactResponse);
+
+    // Extend the reservation to keep it from expire. If the reservation expire,
+    // other tasks can take over the reservation by calling GetOrReserveArtifact.
```
EngHabu, line 57:
```suggestion
    // Extend the reservation to keep it from expiring. If the reservation expires,
    // other tasks can take over the reservation by calling GetOrReserveArtifact.
```

protos/flyteidl/datacatalog/datacatalog.proto
```diff
@@ -33,6 +34,28 @@ service DataCatalog {
...
+    rpc ExtendReservation (ExtendReservationRequest) returns (ExtendReservationResponse);
```
EngHabu, line 58: Should we also add a ReleaseReservation() api? for now the implementation may just return success... but it'll allow the execution engine (propeller) to "cleanup" reservations if the task fails... (optimistically).. even if it fails to do so, the reservation will still expire.. so it's not a correctness fix but more of an optimization so other tasks can start sooner...

CC @kumare3 
```suggestion
    rpc ExtendReservation (ExtendReservationRequest) returns (ExtendReservationResponse);
    
    rpc ReleaseReservation (ReleaseReservationRequest) returns (ReleaseReservationResponse);
```

protos/flyteidl/datacatalog/datacatalog.proto
```diff
@@ -33,6 +34,28 @@ service DataCatalog {
...
     // Return a paginated list of datasets
     rpc ListDatasets (ListDatasetsRequest) returns (ListDatasetsResponse);
+
+    // Get an artifact and the corresponding data. If the artifact does not exist,
+    // try to reserve a spot for populating the artifact.
+    //
+    // Once you preserve a spot, you should call ExtendReservation API periodically
+    // to extend the reservation. Otherwise, the reservation can expire and other
+    // tasks may take the spot.
+    //
```
EngHabu, line 44:
```suggestion
    //
    // If the same owner_id calls this API for the same dataset and it has an active reservation and the artifacts have not been written yet by a different owner, the API will respond with an Acquired Reservation Status (providing idempotency).
    //
```

protos/flyteidl/datacatalog/datacatalog.proto
```diff
@@ -150,6 +173,47 @@ message ListDatasetsResponse {
...
+}
+
+// Whether we successfully reserve a spot.
+message ReservationStatus {
+    enum State {
+        // Acquired the reservation successfully.
+        ACQUIRED = 0;
+
+        // Someone else already took reservation
+        // and it is not available to you.
```
EngHabu, line 190:
```suggestion
        // Indicates an existing active reservation exist for a different owner_id.
```

EngHabu (commented):
A few grammatical nits and a question.
CC @kumare3.


CC @SandraGH5, mind reviewing the inline docs on the APIs (only this file: protos/flyteidl/datacatalog/datacatalog.proto). These end up in our [API Reference docs](https://docs.flyte.org/projects/flyteidl/en/latest/gen/pb-protodoc/flyteidl/datacatalog/datacatalog.proto.html).

protos/flyteidl/datacatalog/datacatalog.proto
SandraGH5: Note: If task A (not the task A)

protos/flyteidl/datacatalog/datacatalog.proto
```diff
@@ -33,6 +34,30 @@ service DataCatalog {
...
+    // tasks may take the spot.
+    //
+    // If the same owner_id calls this API for the same dataset and it has an active reservation and the artifacts have not been written yet by a different owner, the API will respond with an Acquired Reservation Status (providing idempotency).
+    //
+    // Note: We may have multiple concurrent tasks with the same signature
+    // and the same input that try to populate the same artifact at the same time.
+    // Thus with reservation, only one task can run at a time, until the reservation
+    // expires.
+    //
+    // Note: If the task A does not extend the reservation in time and the reservation
```
SandraGH5, line 52: Note: If task A

protos/flyteidl/datacatalog/datacatalog.proto
```diff
@@ -33,6 +34,30 @@ service DataCatalog {
...
+    // expires, another task B may take over the reservation and now we have two tasks
```
SandraGH5, line 53: take over the reservation, resulting in two tasks

protos/flyteidl/datacatalog/datacatalog.proto
```diff
@@ -33,6 +34,28 @@ service DataCatalog {
...
+    // to extend the reservation. Otherwise, the reservation can expire and other
+    // tasks may take the spot.
+    //
+    // Note: We may have multiple concurrent tasks with the same signature
+    // and the same input that try to populate the same artifact at the same time.
+    // Thus with reservation, only one task can run at a time, until the reservation
+    // expire.
+    //
+    // Note: If the task A does not extend the reservation in time and the reservation
+    // may expire. Another task B may take over the reservation and now we have two tasks
```
SandraGH5: reservation, resulting in two tasks

gen/pb_python/flyteidl/datacatalog/datacatalog_pb2_grpc.py
```diff
@@ -112,6 +122,38 @@ def ListDatasets(self, request, context):
...
+    tasks may take the spot.
+
+    If the same owner_id calls this API for the same dataset and it has an active reservation and the artifacts have not been written yet by a different owner, the API will respond with an Acquired Reservation Status (providing idempotency).
+
+    Note: We may have multiple concurrent tasks with the same signature
+    and the same input that try to populate the same artifact at the same time.
+    Thus with reservation, only one task can run at a time, until the reservation
+    expires.
+
+    Note: If the task A does not extend the reservation in time and the reservation
```
SandraGH5, line 140: Note: If task A does not extend the reservation in time and the reservation

gen/pb_python/flyteidl/datacatalog/datacatalog_pb2_grpc.py
```diff
@@ -112,6 +122,38 @@ def ListDatasets(self, request, context):
...
+    expires, another task B may take over the reservation and now we have two tasks
```
SandraGH5, line 141: expires, another task B may take over the reservation, resulting in two tasks

protos/flyteidl/datacatalog/datacatalog.proto
milton0825: Yeah I like the idea.

milton0825: added

milton0825:
Ready for another look

EngHabu (dismissed):
@kumare3 can you review and give +1 as well?

protos/flyteidl/datacatalog/datacatalog.proto
```diff
@@ -150,6 +179,56 @@ message ListDatasetsResponse {
...
+
+// Response to get artifact or reserve spot.
+message GetOrReserveArtifactResponse {
+    oneof value {
+        Artifact artifact = 1;
+        ReservationStatus reservation_status = 2;
+    }
+}
+
+message ExtendReservationRequest {
```
kumare3, line 212: lets add some documentation here as well

protos/flyteidl/datacatalog/datacatalog.proto
```diff
@@ -150,6 +179,56 @@ message ListDatasetsResponse {
...
+    DatasetID dataset_id = 1;
+    string tag_name = 2;
+    string owner_id = 3;
+}
+
+message ExtendReservationResponse {
+
+}
+
+message ReleaseReservationRequest {
```
kumare3, line 222: thank you for adding this.

kumare3:
looks good to me

protos/flyteidl/datacatalog/datacatalog.proto
milton0825: added

protos/flyteidl/datacatalog/datacatalog.proto
milton0825: added

kumare3 (approved):

EngHabu (approved):

PR #146, opened 2021-04-17
Add node execution event version

katrogan:

# TL;DR
Add node execution event version

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/831

## Follow-up issue
_NA_


protos/flyteidl/event/event.proto
```diff
@@ -80,6 +80,11 @@ message NodeExecutionEvent {
 
     // Friendly readable name for the node
     string node_name = 13;
+
+    // Some phases, like RUNNING, can send multiple events with changed metadata (dynamic workflow closures, etc)
+    // that should be recorded regardless of the lack of phase change.
+    // The version field should be incremented when target metadata or other values change across the duration of an individual phase.
+    uint32 phase_version                  = 15;
```
kumare3, line 87: tbh, this seems odd. I realize we have this in Task phases, but I really dont want to add this to node phases :(. It seems like we are not modeling some phase correctly.

Is this because a dynamic node, can get spec, once the parent node completes. If so should we modeling this as a state - DYNAMIC_PARENT_RUNNING, DYNAMIC_PARENT_COMPLETE, CHILD_RUNNING? I know this is hard to do, but the phase versions can be confusing sometimes

PR #159, opened 2021-04-30
Fixed bug with refresh token flow

pmahindrakar-oss:

# TL;DR
Fixes the bug with refresh token flow.

The following comments describe the issue.

 We need to get access tokens expiry as the Expiry on the oauth2.token is refresh Token expiry which is longer.
 This is inorder to solve the following scenario.
 1] Client requests access token and refresh token both in the same request
 2] Auth server issues:
 a) access token with expiry of 30 mins
 b) refresh token with expiry of 60 mins
 3] The combined oath2 token has the expiry for refresh token. ( Alternative is to update the Expiry field with access token expiry)
```
	      {
			access_token 	: "acces_token...."
			refresh_token 	: "refresh_toke..."
			expiry 			: <refresh_token_expiry"
	      }
```
 4] When the check happens on the expiry it will assume access token is valid whereas it may not be and
	      admin server will throw an error that the token is expired.

Solution here is to check the access token expiry and use the refreshTimedelta on this expiry instead of oauth2.Token.Expiry

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [X] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_

### DIFF
--- a/clients/go/admin/pkce/auth_flow_orchestrator.go
+++ b/clients/go/admin/pkce/auth_flow_orchestrator.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/pkg/browser"
 	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/jws"
 )
 
 // TokenOrchestrator implements the main logic to initiate Pkce flow to issue access token and refresh token as well as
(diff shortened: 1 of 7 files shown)


clients/go/admin/pkce/auth_flow_orchestrator.go
```diff
@@ -58,12 +53,38 @@ func (f TokenOrchestrator) FetchTokenFromCacheOrRefreshIt(ctx context.Context) (
...
+	   This is inorder to solve the following scenario.
+	   1] Client requests access token and refresh token both in the same request
+	   2] Auth server issue:
+		  a) access token with expiry of 30 mins
+	      b) refresh token with expiry of 60 mins
+	   3] The combined access token has the expiry for refresh token
+	      {
+			access_token 	: "acces_token...."
+			refresh_token 	: "refresh_toke..."
+			expiry 			: <refresh_token_expiry"
```
EngHabu, line 66: The spec actually specifies that expires_in should be that of the accesstoken. Have you tested with Okta? is that what you see as well? if this is what you see when you setup Google Idp then it's admin's code (in our control)...

Ref: https://tools.ietf.org/html/rfc6749#section-5.1
```
expires_in
         RECOMMENDED.  The lifetime in seconds of the access token.
```

pmahindrakar-oss: Opps. I was skeptical of making this change and thanks for pointing that out from RFC. This was the bug which lurking on admin side and causing issues earlier aswell but was hidden due to earlier logic of force refresh.

pmahindrakar-oss:
Will reopen if  there is someother issue in refresh token flow

PR #169, opened 2021-05-23
Using RST for documentation generation

pmahindrakar-oss:
# TL;DR
* Using protoc gen doc RST support to make it simpler to generate docs and correctly link them internally
  This avoids the post build script to fix the html links which was proposed and implemented here 
  https://github.com/flyteorg/flyteidl/pull/168
* Fixed the documentation warning that arose of the usage of RST which required fixing the comments in the proto files

* As part of this did clean up the proto files to remove the commented out validation code which was using `protoc-gen-validate` . Once we officially support then the code can revived from older version instead of keeping commented out code in the repo

* Also removed unused imports from the proto files which would generate warnings during the doc generation process

* Fixed the referenced links to google.protobuf.[timestamp|Duration|struct] by generating there documentation as part of the core.rst . This doc is generated at the tail end of the core documentation and helps with not having dead links to these where they are referenced in our documentation

* Using a template for all RST file generation except for core. This was done to exclude the Scalar value doc to be generated in each module of the documentation
 https://github.com/pseudomuto/protoc-gen-doc/blob/ae63e7aba9ef91669b2bc76634b33feead0e4f1a/resources/markdown.tmpl#L98
 
 The new template removes the Scalar value and is only generated during core module generation and hence other modules can refer to the scalar values from the core.

*  Updated the contribution doc to use the protoc-gen-doc branch until RST support is officially available

https://docs.flyte.org/projects/flyteidl/en/docs-staging/developing.html#docs-generation

Rest of the docs can be seen here
 https://docs.flyte.org/projects/flyteidl/en/docs-staging/protos/index.html

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


developing.rst
```diff
@@ -18,9 +26,65 @@ Generate Code from protobuf
...
+Relies on protoc plugin pseudomuto/protoc-gen-doc for generating the docs.
+Currently pseudomuto/protoc-gen-doc is being used for generating the RST files which are placed in protos/docs and corresponding folder based on the module.
+Until this code is checkedin https://github.com/pseudomuto/protoc-gen-doc/pull/440, use the branch to build the protoc-gen-doc executable
+Follow the steps to build and use the protoc-gen-doc
+
+#. Clone the git repo protoc-gen-doc
+
+   .. prompt:: bash
+
+     git clone https://github.com/paweld2/protoc-gen-doc.git
```
kumare3, line 60: Why do we need to do this? We can just reference the gitsha in go get right? Other option is to fork the repo into flyteorg and use the fork?

kumare3 (commented):
I would like if the make generate just works without the additional steps

developing.rst
kumare3: Or branch

pmahindrakar-oss:
@kumare3 we need those additional steps for doc generation
We can separate steps for doc generation completely from make generate .

developing.rst
pmahindrakar-oss: Done. Using boiler plate code now and updated the documentation. 
Also currently the gihub workflows don't call make generate and hence don't find any issues. 
We will have to fix the github workflows to use it so that the DELTA_CHECK catches any new files

kumare3 (approved):

wild-endeavor (approved):

PR #176, opened 2021-05-29
Added support for Generic type for extracting a literal value

pmahindrakar-oss:

# TL;DR
Inorder to support generic types, they must be passed in map[string]interface.
This change is required to support generic values to be passed from flytectl eg : DataClass.

Tested example  : 
```
@task
def add(x: Datum, y: Datum) -> Datum:
    """
    Flytekit will automatically convert the passed in json into a DataClass. If the structures dont match, it will raise
    a runtime failure
    """
    x.z.update(y.z)
    return Datum(x=x.x + y.x, y=x.y + y.y, z=x.z)
```

Input to flytectl config:
```
iamRoleARN: "<iamrole>"
inputs:
  "x":
    "x": 2
    "y": ydatafory
    "z":
      1 : "foo"
      2 : "bar"
  "y":
    "x": 3
    "y": ydataforx
    "z":
      3 : "buzz"
      4 : "lightyear"
kubeServiceAcct: ""
targetDomain: ""
targetProject: ""
task: core.type_system.custom_objects.add
version: v3

```

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [ ] Smoke tested
 - [X] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1055

## Follow-up issue
_NA_

### DIFF
--- a/clients/go/coreutils/extract_literal_test.go
+++ b/clients/go/coreutils/extract_literal_test.go
@@ -106,7 +106,10 @@ func TestFetchLiteral(t *testing.T) {
 	})
 
 	t.Run("Generic", func(t *testing.T) {
-		literalVal := "{\"x\": 1, \"y\":\"ystringvalue\"}"
+		literalVal := map[string]interface{}{
+			"x": 1,
+			"y": "ystringvalue",
+		}
 		var literalType = &core.LiteralType{Type: &core.LiteralType_Simple{Simple: core.SimpleType_STRUCT}}
 		lit, err := MakeLiteralForType(literalType, literalVal)
 		assert.NoError(t, err)
--- a/clients/go/coreutils/extract_literal_test.go
+++ b/clients/go/coreutils/extract_literal_test.go
@@ -132,4 +132,29 @@ func TestFetchLiteral(t *testing.T) {
 			assert.Equal(t, val.Kind, extractedStructValue.Fields[key].Kind)
 		}
 	})
+
+	t.Run("Generic Passed As String", func(t *testing.T) {
+		literalVal := "{\"x\": 1,\"y\": \"ystringvalue\"}"
+		var literalType = &core.LiteralType{Type: &core.LiteralType_Simple{Simple: core.SimpleType_STRUCT}}
+		lit, err := MakeLiteralForType(literalType, literalVal)
+		assert.NoError(t, err)
+		extractedLiteralVal, err := ExtractFromLiteral(lit)
+		assert.NoError(t, err)
+		fieldsMap := map[string]*structpb.Value{
+			"x": {
+				Kind: &structpb.Value_NumberValue{NumberValue: 1},
+			},
+			"y": {
+				Kind: &structpb.Value_StringValue{StringValue: "ystringvalue"},
+			},
+		}
+		expectedStructVal := &structpb.Struct{
+			Fields: fieldsMap,
+		}
+		extractedStructValue := extractedLiteralVal.(*structpb.Struct)
+		assert.Equal(t, len(expectedStructVal.Fields), len(extractedStructValue.Fields))
+		for key, val := range expectedStructVal.Fields {
+			assert.Equal(t, val.Kind, extractedStructValue.Fields[key].Kind)
+		}
+	})
 }
--- a/clients/go/coreutils/extract_literal_test.go
+++ b/clients/go/coreutils/extract_literal_test.go
@@ -6,6 +6,9 @@ package coreutils
 import (
 	"testing"
 
+	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/core"
+
+	structpb "github.com/golang/protobuf/ptypes/struct"
 	"github.com/stretchr/testify/assert"
 )
 
@@ -100,5 +103,30 @@ func TestFetchLiteral(t *testing.T) {
 		assert.NotNil(t, p.GetScalar())
 		_, err = ExtractFromLiteral(p)
 		assert.NotNil(t, err)
+	})
+
+	t.Run("Generic", func(t *testing.T) {
+		literalVal := "{\"x\": 1, \"y\":\"ystringvalue\"}"
+		var literalType = &core.LiteralType{Type: &core.LiteralType_Simple{Simple: core.SimpleType_STRUCT}}
+		lit, err := MakeLiteralForType(literalType, literalVal)
+		assert.NoError(t, err)
+		extractedLiteralVal, err := ExtractFromLiteral(lit)
+		assert.NoError(t, err)
+		fieldsMap := map[string]*structpb.Value{
+			"x": {
+				Kind: &structpb.Value_NumberValue{NumberValue: 1},
+			},
+			"y": {
+				Kind: &structpb.Value_StringValue{StringValue: "ystringvalue"},
+			},
+		}
+		expectedStructVal := &structpb.Struct{
+			Fields: fieldsMap,
+		}
+		extractedStructValue := extractedLiteralVal.(*structpb.Struct)
+		assert.Equal(t, len(expectedStructVal.Fields), len(extractedStructValue.Fields))
+		for key, val := range expectedStructVal.Fields {
+			assert.Equal(t, val.Kind, extractedStructValue.Fields[key].Kind)
+		}
 	})
 }
--- a/clients/go/coreutils/literals.go
+++ b/clients/go/coreutils/literals.go
@@ -11,9 +11,9 @@ import (
 	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/core"
 	"github.com/flyteorg/flytestdlib/storage"
 
-	"github.com/golang/protobuf/jsonpb"
 	"github.com/golang/protobuf/ptypes"
 	structpb "github.com/golang/protobuf/ptypes/struct"
+	gateway "github.com/grpc-ecosystem/grpc-gateway/runtime"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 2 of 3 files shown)


clients/go/coreutils/literals.go
```diff
@@ -483,7 +486,20 @@ func MakeLiteralForType(t *core.LiteralType, v interface{}) (*core.Literal, erro
 		}
 	case *core.LiteralType_Simple:
 		newT := t.Type.(*core.LiteralType_Simple)
-		lv, err := MakeLiteralForSimpleType(newT.Simple, fmt.Sprintf("%v", v))
+		strValue := fmt.Sprintf("%v", v)
+		if newT.Simple == core.SimpleType_STRUCT {
+			if _, isValueStringType := v.(string); !isValueStringType {
+				// Using grpc gateway probobuf marshaller which allows marshalling non-proto fields.
```
kumare3, line 492: why?
why not just use simple

json.marshal()?

clients/go/coreutils/literals.go
```diff
@@ -483,7 +484,17 @@ func MakeLiteralForType(t *core.LiteralType, v interface{}) (*core.Literal, erro
 		}
 	case *core.LiteralType_Simple:
 		newT := t.Type.(*core.LiteralType_Simple)
-		lv, err := MakeLiteralForSimpleType(newT.Simple, fmt.Sprintf("%v", v))
+		strValue := fmt.Sprintf("%v", v)
+		if newT.Simple == core.SimpleType_STRUCT {
+			if _, isValueStringType := v.(string); !isValueStringType {
+				byteValue, err := json.Marshal(v)
+				if err != nil {
+					return nil, fmt.Errorf("unable to marshal to yaml string for struct value %v", v)
```
slai, line 492:
```suggestion
					return nil, fmt.Errorf("unable to marshal to JSON string for struct value %v", v)
```

JSON and YAML have different semantics, so worth being specific. If the input from elsewhere, e.g. flytectl, is YAML originally, then this error should be interpreted there instead. Might be worth returning a specific error or error struct instead here.

clients/go/coreutils/extract_literal_test.go
```diff
@@ -101,4 +104,32 @@ func TestFetchLiteral(t *testing.T) {
...
+
+	t.Run("Generic", func(t *testing.T) {
+		literalVal := map[string]interface{}{
+			"x": 1,
+			"y": "ystringvalue",
+		}
+		var literalType = &core.LiteralType{Type: &core.LiteralType_Simple{Simple: core.SimpleType_STRUCT}}
+		lit, err := MakeLiteralForType(literalType, literalVal)
+		assert.NoError(t, err)
+		extractedLiteralVal, err := ExtractFromLiteral(lit)
```
slai, line 116: Is there a reason why this returns a structpb.Struct, instead of a map? It seems asymmetrical to provide a map in, and receive a structpb.Struct on the other side of the roundtrip.

clients/go/coreutils/literals.go
```diff
@@ -483,7 +484,17 @@ func MakeLiteralForType(t *core.LiteralType, v interface{}) (*core.Literal, erro
 		}
 	case *core.LiteralType_Simple:
 		newT := t.Type.(*core.LiteralType_Simple)
-		lv, err := MakeLiteralForSimpleType(newT.Simple, fmt.Sprintf("%v", v))
+		strValue := fmt.Sprintf("%v", v)
+		if newT.Simple == core.SimpleType_STRUCT {
+			if _, isValueStringType := v.(string); !isValueStringType {
```
slai, line 489: Just for my own learning - if it is of SimpleType_STRUCT, why could it be a string?

slai (commented):
I've read what I can from this PR and the linked PRs, and have left some usability comments. 

From what I understand of the use case, the dev story seems a bit unfinished in that you put in a map, and you get a structpb back out.

Also the serialize to and deserialize from JSON just to maintain the function signature will also change the types implicitly, e.g. ints all become float64s, even though protobuf natively supports ints.

kumare3:
> I've read what I can from this PR and the linked PRs, and have left some usability comments. 
> 
> 
> 
> From what I understand of the use case, the dev story seems a bit unfinished in that you put in a map, and you get a structpb back out.
> 
> 
> 
> Also the serialize to and deserialize from JSON just to maintain the function signature will also change the types implicitly, e.g. ints all become float64s, even though protobuf natively supports ints.

You are right @slai  but, generic type is only used for user json. From a user POV this is the vehicle we use for their dataclasses to be transported

clients/go/coreutils/extract_literal_test.go
slai: Given the end goal is to support a DataClass as an input, is there a stronger type that can be used instead of Generic, like some kind of mapping type?

(Sorry, not totally across the Flyte type system :) )

pmahindrakar-oss:
@slai thanks for the review and appreciate your feedback on this.

All execution requests created through the GRPC api  and data is sent over the wire in proto format. 

The execution creation API
https://github.com/flyteorg/flyteidl/blob/419bb2bc913f757b7ec779226925b43fab7fd3e6/gen/pb-go/flyteidl/service/admin.pb.go#L247

And the ExecutionCreateRequest object containing the Inputs for the task/workflow which basicially a LiteralMap

https://github.com/flyteorg/flyteidl/blob/419bb2bc913f757b7ec779226925b43fab7fd3e6/gen/pb-go/flyteidl/admin/execution.pb.go#L81

Now coming to the data type which flyte doesn't know about i think  Generic type is good option where it can just pass over the data in a protobuf container to the task/workflow.

Its represented in json and converted to a proto struct
This is best alternative available in proto package https://pkg.go.dev/google.golang.org/protobuf/types/known/structpb

Client can always do manage this by accepting it as string and deserialising  it on the task and workflow level , but the alternative  of Generic hides that complexity. The client can just define the data class and write his task . When executing it , he will provide a yaml representation of the input data class and flyte will take care of carrying it to the container task/workflow in proto format. 

Lets say we have task which takes both DataClass and integer as inputs , we will convert the first parameter as Generic and second It will make a Primitive lIteral  and store both in the map, serialize these protobufs and send to admin for execution.

clients/go/coreutils/literals.go
pmahindrakar-oss: MakeLiteralForType(t *core.LiteralType, v interface{})  accepts the type and value as an interface and hence its possible for the client of this function to pass in the value as json string directly instead of this function taking care of it. If the string is not in a proper format , it will complain during deserialization.

I have added a unit test `Generic Passed As String`  to explain this .

clients/go/coreutils/literals.go
pmahindrakar-oss: Thanks for pointing that out. Yes this is just json marshalling issue

clients/go/coreutils/extract_literal_test.go
pmahindrakar-oss: Please let me if the response here helps https://github.com/flyteorg/flyteidl/pull/176#issuecomment-853766895

slai:
@pmahindrakar-oss thanks for the additional context.

1. Is there any existing usage of SimpleType_STRUCT? It looks like in this PR, we're constraining it to JSON semantics for all usage, when in the IDL definition, it actually supports a superset of JSON semantics (structpb allows other number types, and I think also supports timestamppb, among other things).

Why not allow _all_ structpb semantics by converting the native Go map to structpb directly, rather than going via JSON?

The current use case might be JSON only, but there's no need to enforce this constraint.

2. I still think ExtractFromLiteral should return a native Go map instead of structpb.Struct because I feel most use cases would require native types. That said, there's precedence for returning pb types in this function anyway, e.g. Timestamp and Duration, so let's keep this behaviour.

clients/go/coreutils/extract_literal_test.go
slai: I still think ExtractFromLiteral should return a native Go map instead of structpb.Struct because I feel most use cases would require native types. That said, there's precedence for returning pb types in this function anyway, e.g. Timestamp and Duration, so let's keep this behaviour.

pmahindrakar-oss:
@slai  i have implemented your suggestion here https://github.com/flyteorg/flyteidl/pull/180

I will have @kumare3 comment on this.  Flytectl is reading into the map[string] interface from a yaml  file.

slai:
@pmahindrakar-oss I've just had a closer look at the definition of [struct.proto](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/struct.proto) and it looks like its semantics are the same as JSON, and it _does not_ support additional number types!

Therefore, there's a little extra work done serializing to JSON and back, but there's no difference in functionality (as far as I know).

Sorry for making you do extra work!

slai (approved):

kumare3 (approved):

pmahindrakar-oss:
Thanks @slai .not a problem. Helped me discover more about structpb.

clients/go/coreutils/literals.go
EngHabu, line 489: if v is a string, why do we try to json marshal it? won't that escape characters and cause havoc?

v can also be a map[string]interface{}, right? or just any type really... that we should json.Marshal... and json unmarshal into a pb struct.. right?

katrogan (approved):

clients/go/coreutils/literals.go
pmahindrakar-oss: Hi @EngHabu , we dont actually do json  marshal if its a string .  the condition is ! isValueStringType so we skip over if its a string  and pass it as it is to jsonpb.Unmarshal to unmarshal it to a structpb.

And yes v can be map[string]interface{} , which is what we are using from flytectl and that one will be json Marshalled and unmarshalled to pb struct.

pmahindrakar-oss: Check this test `Generic Passed As String` https://github.com/flyteorg/flyteidl/pull/176/files/39b9809db14ba0fd508958d72534580d4a453dc9#diff-87d3befa5f3d7a578349048f3e4711bcaee22606ce57a9339a8fd910e27e7168R136

And this one which passes map[string]interface 
https://github.com/flyteorg/flyteidl/pull/176/files/39b9809db14ba0fd508958d72534580d4a453dc9#diff-87d3befa5f3d7a578349048f3e4711bcaee22606ce57a9339a8fd910e27e7168R109

Any other type or invalid string will cause failure during unmarshal which we relying for throwing an error.

EngHabu: 👍🏾

EngHabu (approved):

PR #178, opened 2021-06-01
Proposal: Introduce Array Nodes

katrogan:

# TL;DR
In the case of map tasks lets define the _parent task_ as the maptask definition in user code and the _subtask_ as what it yields. The task container target is insufficient to fully encapsulate the subtask because:

- both the parent task and subtask read from shared TaskTemplate fields such as the custom and task template version
- to support mapped pod tasks, the parent task needs to now construct spec from the K8sPod target alongside TaskTemplate fields

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1051
https://github.com/flyteorg/flyte/issues/1131

## Follow-up issue
_NA_


katrogan:
PTAL @wild-endeavor @EngHabu

protos/flyteidl/core/workflow.proto
```diff
@@ -63,6 +63,26 @@ message WorkflowNode {
...
 
+// Configures an array job in which the node executes a task in parallel across a range of inputs.
+message ArrayNode {
+    // Reference to a registered task which should be called in parallel during node execution.
+    Identifier task_reference = 1;
+
+    // Defines the minimum number of instances to bring up concurrently at any given point. Note that this is an
+    // optimistic restriction and that, due to network partitioning or other failures, the actual number of currently
+    // running instances might be more. This has to be a positive number if assigned. Default value is size.
+    int64 parallelism = 2;
```
kumare3, line 74: I would call this concurrency

katrogan: done

PR #183, opened 2021-06-04
#minor Enum type

kumare3:
# TL;DR
Supports new EnumType in Flyte type system

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Only string enum types are supported. this is purely a type system update, the underlying transport is still a stringvalue

## Tracking Issue
https://github.com/flyteorg/flyte/issues/590


protos/flyteidl/core/types.proto
```diff
@@ -73,6 +81,9 @@ message LiteralType {
 
         // A blob might have specialized implementation details depending on associated metadata.
         BlobType blob = 5;
+
+        // Defines an enum with pre-defined string values.
+        EnumType enum_type = 7;
```
EngHabu, line 86: Should it be under Simple?

kumare3: Nope, Simple types are just enums (i.e) they are declared only using an enum value. While enums have values associated with them. The transport is actually just a string

katrogan (approved):

EngHabu (approved):
Looks good.

katrogan (approved):

protos/flyteidl/core/types.proto
```diff
@@ -56,6 +56,14 @@ message BlobType {
     BlobDimensionality dimensionality = 2;
 }
 
+// Enables declaring enum types, with predefined string values
+message EnumType {
+    // Predefined set of enum values
+    repeated string values = 1;
+    // default value for the enum type
+    string default_value = 2;
```
EngHabu, line 64: Should the default value just be the first value?

EngHabu (changes requested):

protos/flyteidl/core/types.proto
kumare3: I wanted to have a way of not having a default. Thus in the drop down no value could be selected

EngHabu: I actually like the protobuf decision and golang's of having the first be the default and forcing people, if they want, to name the first one UNDEFINED... 

I think it also simplifies UI implementation and other sdk languages...

kumare3: I actually think protobuf's decision was based on the wire-format and that `0` value is the default int value.
I am thinking of cases, where you do not want a value to be selected at all. I understand we can define a `UNDEFINED` value, but this for users - who never use protobuf (think data scientists, data engineers) and users who use python - which supports None value, would be harder to understand.

But I am ok with whatever, as I think we can create the representation in higher levels. Just that ever ux layer wil need to represent it in the same way

kumare3: Let me just remove the default value, we can always add that in

kumare3:
@EngHabu / @katrogan +1 again please

EngHabu (approved):

katrogan (approved):

PR #184, opened 2021-06-05
[wip] #minor offloadable Dense vectors

kumare3:
# TL;DR
Support for dense vectors in flyteidl. Instead of using multivariate LiteralColletion, Flyteidl now supports offloadable dense vectors for certain types. This is interesting for may ML type of operations

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/<number>

## Follow-up issue
_NA_
OR


protos/flyteidl/core/literals.proto
```diff
@@ -75,15 +75,80 @@ message Literal {
...
+        // DenseVector of type float (32bit precision)
+        FloatVector float_vector = 6;
+        // DenseVector of type Double (64bit precision)
+        DoubleVector double_vector = 7;
+        // DenseVector of type Bool
+        BoolVector bool_vector = 8;
+        // DenseVector of type string
+        StringVector str_vector = 9;
     }
 }
```
EngHabu, line 95: I think `Offloaded` should be added here instead... a Generic/String Scalar can be arbitrary large that we want it offloaded, same for Literal Collection/Map...

```suggestion
        StringVector str_vector = 9;
    }
    
    Offloaded offloaded = 10;
}
```

kumare3: ya I was thinking about that, the reason, why I did not add it here, is because we will then face major `TypeErasure` and lose the type information at runtime. I initially create a singular DenseVector which can be offloaded and one of the typed vectors. But, then if the value is offloaded at runtime, it is very hard to know the type. This has been causing some problems - i.e., we will always need to load the relevant interface.

In the approach I have taken, the shell is maintained, thus type is implicitly derivable

EngHabu: If maintaining the type is a goal for offloading, we can add the LiteralType to the offload message...

If we follow the approach in this PR, how will we ever offload the literalcollection for example? It'll be impossible to while maintaining the type... Right?

protos/flyteidl/core/literals.proto
```diff
@@ -75,15 +75,80 @@ message Literal {
...
+message Offloaded {
+    // URI of the offloaded vector. Of the form s3://...
+    string uri = 1;
+    // Format of the offloaded vector. this is for the client libraries to parse. Example: parquet, arrow, hd5, proto etc
+    string format = 2;
+}
+
+// Represents a list of Strings
+message StringVector {
+    repeated string values = 1;
```
katrogan, line 106: why not oneof values or offloaded?

PR #187, opened 2021-06-15
Add endpoint to recover workflow execution

katrogan:

# TL;DR
Add endpoint to recover workflow execution.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1151

## Follow-up issue
_NA_


protos/flyteidl/core/execution.proto
```diff
@@ -18,6 +18,7 @@ message WorkflowExecution {
         FAILED = 6;
         ABORTED = 7;
         TIMED_OUT = 8;
+        REOVERED = 9;
```
wild-endeavor, line 21:
```suggestion
        RECOVERED = 9;
```

protos/flyteidl/core/execution.proto
```diff
@@ -34,6 +35,7 @@ message NodeExecution {
         SKIPPED = 7;
         TIMED_OUT = 8;
         DYNAMIC_RUNNING = 9;
+        REOVERED = 10;
```
wild-endeavor, line 38:
```suggestion
        RECOVERED = 10;
```

protos/flyteidl/admin/execution.proto
```diff
@@ -49,6 +49,21 @@ message ExecutionRelaunchRequest {
...
+    // Identifier of the workflow execution to recover.
+    core.WorkflowExecutionIdentifier id = 1;
+
+    // User provided value for the recovered execution.
+    // If none is provided the system will generate a unique string.
+    // +optional
+    string name = 2;
+
+    // Which subworkflow node (if any) launched this execution
+    core.NodeExecutionIdentifier parent_node_execution = 3;
```
wild-endeavor, line 63: can we add a comment for why this is necessary?  it's not immediately obvious.

protos/flyteidl/core/execution.proto
katrogan: oof, big thanks!

protos/flyteidl/admin/execution.proto
katrogan: done, let me know if this is clearer

wild-endeavor: Not really, but I think it's because I don't understand this recovery feature/what a parent_node_execution is.

protos/flyteidl/service/admin.proto
```diff
@@ -248,6 +248,16 @@ service AdminService {
     };
   }
 
+  rpc RecoverExecution (flyteidl.admin.ExecutionRecoverRequest) returns (flyteidl.admin.ExecutionCreateResponse) {
+    option (google.api.http) = {
+      post: "/api/v1/executions/recover"
+      body: "*"
+    };
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+      description: "Recovers a previously-run workflow execution that failed due to system failure."
```
kumare3, line 257: can we improve the description - 
In Recover mode, users cannot change any input parameters or update the version of the execution. This is extremely useful to recover from system errors and byzantine faults like - Loss of K8s cluster, bugs in platform or instability, machine failures, downstream system failures (downstream services), or simply to recover executions that failed because of retry exhaustion and should complete if tried again.

kumare3: also add a `:ref:`?

protos/flyteidl/admin/execution.proto
kumare3: I agree with Yee, this is confusing! Why do we need to know the parent at all. If you just recover one of the child workflows, then it should simply recover the child workflow! I do not think we should try and link it to the parent - so node execution identifier should not be required?

protos/flyteidl/core/execution.proto
```diff
@@ -34,6 +35,7 @@ message NodeExecution {
         SKIPPED = 7;
         TIMED_OUT = 8;
         DYNAMIC_RUNNING = 9;
+        RECOVERED = 10;
```
kumare3, line 38: we should also update the statemachine diagram to indicate this new state in the docs. Also while you are at it, you can use the fancy sphinx plugin that Yee introduced

gen/pb-go/flyteidl/service/flyteadmin/api/swagger.yaml
```diff
@@ -475,6 +475,25 @@ paths:
           description: "A successful response."
           schema:
             $ref: "#/definitions/adminExecutionCreateResponse"
+  /api/v1/executions/recover:
+    post:
+      tags:
+      - "AdminService"
+      description: "Recovers a previously-run workflow execution that failed due to\
+        \ system failure."
```
pmahindrakar-oss, line 483: Does system failure mean a set of error codes . If yes ,  can we  add those in the description?

gen/pb-go/flyteidl/service/flyteadmin/api/swagger.yaml
```diff
@@ -3652,6 +3673,22 @@ definitions:
...
+    properties:
+      id:
+        description: "Identifier of the workflow execution to recover."
+        $ref: "#/definitions/coreWorkflowExecutionIdentifier"
+      name:
+        type: "string"
+        title: "User provided value for the recovered execution.\nIf none is provided\
+          \ the system will generate a unique string.\n+optional"
+      parent_node_execution:
+        description: "Which node (if any) launched this execution. This situation\
```
pmahindrakar-oss, line 3687: when you mean this -> you mean the referenced execution?

protos/flyteidl/admin/execution.proto
katrogan: This is to mimic the existing behavior when a parent workflow launches a child one. We pass through execution metadata like the principal

protos/flyteidl/core/execution.proto
katrogan: good point, but i will do that in separate PR since it's in the flyte repo

protos/flyteidl/admin/execution.proto
EngHabu: I understand why we need to overwrite this, @kumare3. When you recover Exec1 that has a node that, ran Exec2 and then we failed somewhere in there... when we recover Exec1', and we come to that node, it should launch Exec2' in recover mode... but if all it did was call the Recover endpoint on Exec2, then it'll look like it's associated with the node in Exec1 not Exec1'... 

This sounds like we want to override _somethings_ within the `ExecutionSpec` of the recovered execution... and right now we think we only need to overwrite [parent_node_execution](https://github.com/flyteorg/flyteidl/blob/f4809c1a52073ec80de41be109bccc6331cdbac3/protos/flyteidl/admin/execution.proto#L200).

But I guess we might even want to overwrite [reference_execution](https://github.com/flyteorg/flyteidl/blob/f4809c1a52073ec80de41be109bccc6331cdbac3/protos/flyteidl/admin/execution.proto#L204) later on... so maybe we should have a way to overwrite all `ExecutionMetadata` fields?

protos/flyteidl/core/execution.proto
```diff
@@ -18,6 +18,7 @@ message WorkflowExecution {
         FAILED = 6;
         ABORTED = 7;
         TIMED_OUT = 8;
+        RECOVERED = 9;
```
EngHabu, line 21: What does this state represent? so when we create a new execution, it'll go to RECOVERED instead of QUEUED or before/after QUEUED? I'm not sure we actually need a state to track it... mind clarifying how that will work?

gen/pb-go/flyteidl/service/flyteadmin/api/swagger.yaml
katrogan: updated

protos/flyteidl/core/execution.proto
katrogan: removed, thanks for pointing out

gen/pb-go/flyteidl/service/flyteadmin/api/swagger.yaml
katrogan: just all errors for now

protos/flyteidl/admin/execution.proto
katrogan: fair point @EngHabu do you mean we should pass an Execution Metadata instead here? I do question whether we want to overwrite the reference execution, admin can implicitly do this already based on it being the RecoverExecution endpoint, it seems error-prone to leave it to the caller to set that themselves

katrogan:
PTAL @EngHabu @kumare3

protos/flyteidl/service/admin.proto
kumare3: also add about transient failures

protos/flyteidl/admin/execution.proto
```diff
@@ -178,6 +194,9 @@ message ExecutionMetadata {
 
         // This execution was triggered by another execution.
         CHILD_WORKFLOW = 4;
+
+        // This execution was recovered from another execution.
+        RECOVERED = 5;
```
kumare3, line 199: can it not be recovered and child?

katrogan: good point, updated

protos/flyteidl/admin/execution.proto
EngHabu: I did mean that yes..
What should the reference_execution be? should it be the execution we are recovering from? or the "parent" execution that is sending this recover request?

protos/flyteidl/service/admin.proto
katrogan: so the generated rst docs all pick up the comments above, these descriptions are only used for swagger. I can make them consistent-ish although the rst links don't work.

protos/flyteidl/core/execution.proto
katrogan: sure, this is to indicate that although a node succeeded it wasn't really run, rather it was recovered. if you think this is unnecessary i can just default to success

protos/flyteidl/admin/execution.proto
```diff
@@ -178,6 +194,12 @@ message ExecutionMetadata {
...
+
+        // This execution was a child workflow recovered from another execution.
+        RECOVERED_CHILD_WORKFLOW = 6;
```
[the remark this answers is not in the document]
katrogan: okay, added it on ketan's suggestion. removed

protos/flyteidl/admin/execution.proto
katrogan: in this case the parent. the execution that we're recovering from is the source execution id (stored separately)

katrogan:
PTAL @EngHabu @kumare3

kumare3 (approved):
awesome!

protos/flyteidl/core/execution.proto
EngHabu: aha.. so this is a terminal state that a node will move into when it's been recovered... cool cool.. SG!

EngHabu (approved):

PR #215, opened 2021-08-24
Cache Serialize API

hamersaw:
# TL;DR
Removes the ExtendReservation API from the datacatalog protobuf definition.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
The datacatalog ExtendReservation API is continually called when an Artifact is already reserved and that reservation needs to be extended past the current expiration. We are changing the implementation of the GetOrReserveArtifact API to extend a reservation if valid. Therefore, ExtendReservation is no longer necessary. We have included additional fields into the GetOrReserveArtifactResponse to reflect these changes.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/267
https://github.com/flyteorg/flyte/issues/872

## Follow-up issue
_NA_


protos/flyteidl/core/catalog.proto
```diff
@@ -41,3 +41,17 @@ message CatalogMetadata {
        TaskExecutionIdentifier source_task_execution = 3;
     }
 };
+
+// Indicates the status of a catalog reservation operation.
+enum CatalogReservationStatus {
```
EngHabu, line 46: You may want to move this into a message... because of the way code generation happens, this will end up with weird naming in some languages...

EngHabu (dismissed):
Finally started looking at the cache serialization PRs... 
Let's wait on merging them until they are all +1'd

protos/flyteidl/core/catalog.proto
hamersaw: @EngHabu  We'll want to move the CatalogCacheStatus enum at the top of this file then as well. Do we just want to move all of this file out of the core package and into the datacatalog package?

EngHabu: Example: https://github.com/flyteorg/flyteidl/blob/master/protos/flyteidl/core/execution.proto

I don't think you need to do the refactoring right now... we need a bigger project to revisit DataCatalog design and we can do the cleanup as part of that... but at least for new additions (like this PR), let's prevent future cleanup

hamersaw: sounds great, done.

EngHabu (dismissed):

protos/flyteidl/core/tasks.proto
```diff
@@ -92,6 +92,9 @@ message TaskMetadata {
     oneof interruptible_value {
         bool interruptible = 8;
     };
+
+    // Indicates whether the system should attempt to execute discoverable instances in serial to avoid duplicate work
+    bool discovery_serializable = 9;
```
kumare3, line 97: i hate this name. I guess we want to keep it so.
but, ideally it should be cache_serializable right?

kumare3 (dismissed):
one comment

protos/flyteidl/core/tasks.proto
hamersaw: done. i'm ok with the switch, it hurt me to try and keep it uniform.

EngHabu (approved):

kumare3 (approved):

PR #221, opened 2021-09-10
Update EventError with Is and Unwrap functions

hamersaw:
Signed-off-by: Daniel Rammer <daniel@union.ai>

# TL;DR
The golang errors package provides the functions Is (to check if an instance of the error type is within the stack), As (to convert the error to the error type if it is within the stack), and Unwrap (to facilitate the aforementioned functionality). This updates the EventError type to enable use within the API.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
See TL;DR

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/clients/go/events/errors/errors.go
+++ b/clients/go/events/errors/errors.go
@@ -2,6 +2,7 @@ package errors
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/admin"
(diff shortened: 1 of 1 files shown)


clients/go/events/errors/errors.go
```diff
@@ -31,6 +32,18 @@ func (r EventError) Error() string {
 	return fmt.Sprintf("%s: %s, caused by [%s]", r.Code, r.Message, r.Cause.Error())
 }
 
+func (r *EventError) Is(target error) bool {
+	t, ok := target.(*EventError)
+	if !ok {
+		return false
+	}
+	return r.Code == t.Code && (r.Cause == t.Cause || t.Cause == nil) && (r.Message == t.Message || t.Message == "")
```
EngHabu, line 40: Should that just check for the Code? I feel the message and cause may be different but the errors will still be _the same_... it's not an "Equals" check.. or at least in my opinion it doesn't have to be...

hamersaw: Good point. I saw it implemented this way elsewhere and thought it might add more extensibility. The idea is that when checking for an error if you don't provide the Cause or Message than it performs the check as you suggested - equality over Code and then Cause == nil / Message == "". This scenario is applied in the updates to IsNotFound, IsAlreadyExist, etc below. For example:

`errors.Is(err, &EventError{Code: ExecutionNotFound}`

Only checks if the Code is equal to any error in the stack. However, if a user wanted to include more fine-grained checks they could include a Cause or Message as well. I don't have strong feelings either way, if you prefer to perform only a Code check I can make the update.

EngHabu: SG 👍🏽

clients/go/events/errors/errors.go
```diff
@@ -31,6 +32,18 @@ func (r EventError) Error() string {
 	return fmt.Sprintf("%s: %s, caused by [%s]", r.Code, r.Message, r.Cause.Error())
 }
 
+func (r *EventError) Is(target error) bool {
+	t, ok := target.(*EventError)
+	if !ok {
+		return false
+	}
```
EngHabu, line 39: The one additional check I would add is:
```suggestion
	}
	
	if r == nil && t == nil {
	    return true
	}
	
	if r == nil || t == nil {
	    return false
	}
```

Example for why: https://play.golang.org/p/P85Pext7xw5

hamersaw: Very interesting, it doesn't seem like the check should be necessary, but the example certainly clarifies it.

EngHabu (approved):

wild-endeavor (approved):
is this backwards compatible with everything?

hamersaw:
> is this backwards compatible with everything?

@wild-endeavor Yes, I just added a few functions to support traversal through the entire error stack to check for EventError types.

PR #230, opened 2021-11-17
Add support union type

pingsutw:
Signed-off-by: Kevin Su <pingsutw@apache.org>

# TL;DR
Add union type
https://github.com/flyteorg/flyte/issues/1349
I think we don't need to add an optional type here because `typing.Optional[str]` is equal to `typing.Union[str, None]`

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1349

## Follow-up issue
_NA_


maximsmol:
if we are going with the tagged union implementation, don't we also need to add a new `LiteralRepr`?

see https://github.com/maximsmol/flyte/blob/master/rfc/core%20language/sum-types.md#3-proposed-implementation

pingsutw:
yeah, we do need to add `LiteralRepr` if we are going to add a type checker in the propeller.

PR #235, opened 2021-12-02
Union Types #minor

maximsmol:
# TL;DR
Tagged Union types to support `typing.Union` in Python

Inherited from https://github.com/flyteorg/flyteidl/pull/230

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added
 - [X] Any pending items have an associated Issue

## Complete description
https://github.com/maximsmol/flyte/blob/master/rfc/core%20language/sum-types.md

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1349

## Follow-up issue
_NA_


protos/flyteidl/core/literals.proto
```diff
@@ -52,6 +52,13 @@ message Schema {
     SchemaType type = 3;
 }
 
+// The runtime representation of a tagged union value. See `UnionType` for more details.
+message Union {
+    Literal value = 1;
+    UnionType type = 2;
+    uint64 tag = 3;
```
EngHabu, line 59: Should we replace this with:
```suggestion
    LiteralType type = 2;
```

Or in other words, what's the benefit of doing this vs just stating the LiteralType?

cc @wild-endeavor

maximsmol: 1. Agreed, was originally there to deal with Unions with different orderings in Python (e.g. `Union[str, int]` and `Union[int, str]` would be incompatible if all we had was an integer index), but given [the switch to string tags](https://github.com/maximsmol/flyte/blob/master/rfc/core%20language/sum-types-2.md) this is no longer needed

2, 3, 4. It's a good question whether we should allow using a value as a union implicitly as we would have to guess which instantiation was meant and, in languages like C++, the choice would be arbitrary. For example, how do we instantiate `std::variant<int, int>` i.e. `Tag0 Int | Tag1 Int` with a `LiteralRepr` containing an integer? Do we pick the first or the second variant? This is similarly the case for languages with algebraic data types (symbolically tagged unions). For Python we also have to make an arbitrary choice between two type transformers when instantiating `Union[MyInt | int]` but this case could be solved with somehow passing along the return type's transformer's name from the Python source annotation. This trick would not work for `std::variant`

In any case, if we decided to somehow do this automatic wrapping, we already have the `UnionType` accessible through `validateBinding`'s `expectedType` argument, so I don't see a reason to pass it along. I might be misunderstanding something though.

maximsmol: Thought about it some more, and I think we actually do want to do automatic wrapping iff unambiguous

Example:
```py
def f() -> str:
  return "10"
  
def g(x: Union[int, str]) -> str:
  return str(x)
  
def h(x: Union[MyStr, str]) -> str:
  return str(x)
  
g(f()) # all good, unambiguous
h(f()) # compilation failure

def workaround(x: str) -> Union[MyStr, str]:
  return MyStr(x)
  
h(workaround(f())) # all good, unambiguous (union literal includes disambiguating tag)
```

maximsmol:
closed by mistake (automatically by github after branch rename)

eapolinario (approved):

EngHabu (approved):
Thank you!

PR #244, opened 2022-01-03
Add RetryAttempt and Phase to ExternalResourceInfo proto

hamersaw:
# TL;DR
To display map task execution information in FlyteConsole we need to add the RetryAttempt and Phase to the ExternalResourceInfo proto message to track subtasks.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
^^^

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1838

## Follow-up issue
_NA_


protos/flyteidl/event/event.proto
```diff
@@ -193,6 +193,12 @@ message ExternalResourceInfo {
 
     // Identifier for an external resource created by this task execution, for example Qubole query ID or presto query ids.
     string external_id = 1;
+
+    // Retry attempt number for this external resource, ie., 2 for the second attempt
+    uint32 retry_attempt = 2;
```
EngHabu, line 198: I think we may need to add someway of tying multiple ExternalResourceInfo objects together...
if we have an array job of size 3, it may report 
ex1, ex2, ex3...
then in some future iteration, may report:
ex1, ex2, ex3, ex4

With no way of telling which original subtask is ex4 a retry of...

Maybe we should add an index field too?

hamersaw: @EngHabu  please correct me if I am off on any / all of this.

It looks like each time the array plugin launches new tasks (across both aws batch and k8s) we call [DetermineDiscoverability](https://github.com/flyteorg/flyteplugins/blob/973a51ae8cc6a06e2f78007fe1ffda71f7355fad/go/tasks/plugins/array/catalog.go#L29) to check whether each task is cached or not. This sets the [IndexesToCache](https://github.com/flyteorg/flyteplugins/blob/973a51ae8cc6a06e2f78007fe1ffda71f7355fad/go/tasks/plugins/array/core/state.go#L54) field on the State.

So any time new tasks are launched, or existing instances are checked, the [ExecutionArraySize](https://github.com/flyteorg/flyteplugins/blob/973a51ae8cc6a06e2f78007fe1ffda71f7355fad/go/tasks/plugins/array/core/state.go#L48) is always less than or equal to the [OriginalArraySize](https://github.com/flyteorg/flyteplugins/blob/973a51ae8cc6a06e2f78007fe1ffda71f7355fad/go/tasks/plugins/array/core/state.go#L49).

The [Detailed](https://github.com/flyteorg/flyteplugins/blob/973a51ae8cc6a06e2f78007fe1ffda71f7355fad/go/tasks/plugins/array/arraystatus/status.go#L20) bitarray.CompactArray is used to track the phases of subtasks through each iteration. It is of size "ExecutionArraySize", which because of caching between task executions, may change between iterations. Currently, both the aws batch and k8s array implementations use indices into this array to generate subTaskIDs - which propagate to the external resource IDs. Therefore, the IDs for consecutive executions may not necessarily align with the same tasks.

What I propose is that rather than using the index into the Detailed array (whose size may change between iterations) we use the original index to assign subTaskIDs. This is available using the [CalculateOriginalIndex](https://github.com/flyteorg/flyteplugins/blob/973a51ae8cc6a06e2f78007fe1ffda71f7355fad/go/tasks/plugins/array/core/state.go#L321) function which uses the bitset of cached tasks to determine the original index. Then subTaskIDs (and the externalID in the ExternalResources proto) will stay constant across executions without maintaining a new Index field in the proto message. This makes more sense to me.

Additionally, I think the task logs [currently use](https://github.com/flyteorg/flyteplugins/blob/973a51ae8cc6a06e2f78007fe1ffda71f7355fad/go/tasks/plugins/array/k8s/task.go#L194) the original index. So linking back to all retries of a particular sub task is much easier if the external resource ID matches the log link.

EngHabu (approved):

EngHabu:
Mind rebasing to master so we can merge this?

EngHabu (approved):

katrogan (approved):

PR #248, opened 2022-01-11
Added UpdateExecution API and api objects

pmahindrakar-oss:

# TL;DR

IDL changes for archiving / activating  executions

Testing done :

```
(base) ➜  flytectl git:(pmahindrakar/2045) ✗ flytectl create execution -p flytesnacks -d development --execFile lp.yaml
execution identifier project:"flytesnacks" domain:"development" name:"fb2e78be613d64643a05" 
(base) ➜  flytectl git:(pmahindrakar/2045) ✗ flytectl get executions -p flytesnacks -d development                                                      
 ---------------------- ------------------------------- ------------- ----------- ---------------- -------------------------------- --------------- -------------------- -------------------- 
| NAME                 | LAUNCH PLAN NAME              | TYPE        | PHASE     | SCHEDULED TIME | STARTED                        | ELAPSED TIME  | ABORT DATA (TRUNC) | ERROR DATA (TRUNC) |
 ---------------------- ------------------------------- ------------- ----------- ---------------- -------------------------------- --------------- -------------------- -------------------- 
| fb2e78be613d64643a05 | core.flyte_basics.lp.go_greet | LAUNCH_PLAN | SUCCEEDED |                | 2022-01-11T07:51:38.121980703Z | 25.171435025s |                    |                    |
 ---------------------- ------------------------------- ------------- ----------- ---------------- -------------------------------- --------------- -------------------- -------------------- 
1 rows
(base) ➜  flytectl git:(pmahindrakar/2045) ✗ flytectl update execution -p flytesnacks -d development fb2e78be613d64643a0 --archive 
Execution fb2e78be613d64643a0 failed to get updated due to rpc error: code = NotFound desc = missing entity of type execution with identifier project:"flytesnacks" domain:"development" name:"fb2e78be613d64643a0" 
Error: rpc error: code = NotFound desc = missing entity of type execution with identifier project:"flytesnacks" domain:"development" name:"fb2e78be613d64643a0" 
(base) ➜  flytectl git:(pmahindrakar/2045) ✗ flytectl update execution -p flytesnacks -d development fb2e78be613d64643a05  --archive --activate
Error: Invalid state passed. Specify either activate or archive

(base) ➜  flytectl git:(pmahindrakar/2045) ✗ flytectl update execution -p flytesnacks -d development fb2e78be613d64643a05  --archive           
updated execution successfully on fb2e78be613d64643a05 to state EXECUTION_ARCHIVED%                                                                                                                 (base) ➜  flytectl git:(pmahindrakar/2045) ✗ flytectl update execution -p flytesnacks -d development fb2e78be613d64643a05  --archive 
updated execution successfully on fb2e78be613d64643a05 to state EXECUTION_ARCHIVED%                                                                                                                 (base) ➜  flytectl git:(pmahindrakar/2045) ✗ flytectl update execution -p flytesnacks -d development fb2e78be613d64643a05  --activate
updated execution successfully on fb2e78be613d64643a05 to state EXECUTION_ACTIVE%   

```
## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Complete description

## Tracking Issue
 https://github.com/flyteorg/flyte/issues/2045

## Follow-up issue
_NA_


katrogan (approved):

anrusina (approved):

protos/flyteidl/admin/execution.proto
```diff
@@ -320,3 +322,25 @@ message WorkflowExecutionGetDataResponse {
...
+    // The state of the execution is used to control its visibility in the UI/CLI.
+    enum ExecutionState {
+        // By default, all executions are considered active.
+        EXECUTION_ACTIVE = 0;
+
+        // Archived executions are no longer visible in the UI.
+        EXECUTION_ARCHIVED = 1;
+    }
+    ExecutionState state = 1;
+}
```
EngHabu, line 344: Should we store occured_at here as well?

EngHabu (dismissed):

protos/flyteidl/admin/execution.proto
pmahindrakar-oss: Instead of accepting that from the API request which is basically going to be users timestamp of when they execute flytectl update execution <execid> --archive or whichever other form to call the api.

Do we instead want the time stamp to be generated by the system when the request is receieved and store it in the model as StateUpdatedAt timestamp

I have seen in some other PR's where we are doing this system generated timestamps instead of accepting from request.

Having a consistent way of doing this would be preferable in my opinion for such audit related attributes
Let me know what you think

katrogan (approved):

protos/flyteidl/admin/execution.proto
EngHabu: True... I'm not suggesting to send this in the request. I think it should be generated by the server but I thought we use the same proto for storage?

pmahindrakar-oss: @EngHabu can you elaborate further on which proto you are talking about.  And also if we are not accepting it in the request then do we need any api changes here . We can handle those directly in consumer flyteadmin right with just the model changes

EngHabu: This `message ExecutionStatus {` do we store this in the DB raw? do we know when was it archived? just looking at updated_at seem error prune... 
@katrogan

pmahindrakar-oss: Also my assumption is that this StateUpdatedAt is not exposed in the UI or through CLI. If we do want to show this state then we would need the API change.

katrogan: we shouldn't depend on updated_at because if someone archives an in-progress execution, its termination event will re-set updated_at. @EngHabu just curious, why do we care to store the state change occurred time? could we use the admin audit logs to capture who updated the state and when?

katrogan: @pmahindrakar-oss if we go with your StateUpdatedAt approach won't we need a new protobuf field to expose that, which _will_ require some idl hange?

pmahindrakar-oss: I was thinking it was just for reporting purposes which read directly from the DB but i see if we want queries on this field from UI /ctl then we need to expose it in the API. I have added the field.

katrogan: hey @pmahindrakar-oss we were thinking it might make sense to capture the user in addition to the timestamp when archiving a workflow, like the [principal](https://github.com/flyteorg/flyteidl/blob/master/protos/flyteidl/admin/execution.proto#L209).

However, we shouldn't let necessarily let users actually specify the principal or the timestamp when issuing the archive request. Rather the user/principal should come from the authenticated context like [here](https://github.com/flyteorg/flyteadmin/blob/master/pkg/manager/impl/execution_manager.go#L750)

So perhaps we could have two separate messages, one that only contains the execution state and that is included in the body of the UpdateExecutionRequest message. Then we would add a separate message containing
-ExecutionState
-OccurredAt
-Principal

which is embedded in the closure and that gets returned for Get/List execution requests

pmahindrakar-oss: I have updated the IDL , please have a look

katrogan (approved):
looks good, thank you for all the changes!

EngHabu (approved):

PR #262, opened 2022-03-17
Add custom token source that allows preemptive token refresh

mayitbeegh:
# TL;DR
Currently the OAuth token gets refreshed once the token expires. When 1000 flytepropeller pods start up around the same time, they DDOS the auth server every time their tokens expire. This PR tries to address this problem by adding a custom token source that preemptively refreshes the token at a random time before the token expires and after the configured time limit. 

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Follow-up issue
_NA_


clients/go/admin/config.go
```diff
@@ -37,14 +37,16 @@ const (
...
+	InsecureSkipVerify        bool            `json:"insecureSkipVerify" pflag:",InsecureSkipVerify controls whether a client verifies the server's certificate chain and host name. Caution : shouldn't be use for production usecases'"`
+	CACertFilePath            string          `json:"caCertFilePath" pflag:",Use specified certificate file to verify the admin server peer."`
+	MaxBackoffDelay           config.Duration `json:"maxBackoffDelay" pflag:",Max delay for grpc backoff"`
+	PerRetryTimeout           config.Duration `json:"perRetryTimeout" pflag:",gRPC per retry timeout"`
+	MaxRetries                int             `json:"maxRetries" pflag:",Max number of gRPC retries"`
+	AuthType                  AuthType        `json:"authType" pflag:"-,Type of OAuth2 flow used for communicating with admin."`
+	RefreshTokenPreemptively  bool            `json:"refreshTokenPreemptively" pflag:",Preemptively refreshes auth token with jitter before it expires (client credential only)."`
+	PreemptiveRefreshDuration config.Duration `json:"preemptiveRefreshDuration" pflag:",Max duration between token refresh attempt and token expiry."`
```
EngHabu, line 49: Can we change this to `TokenRefreshWindow` (or a better babe).. so if it's 0, it means we will refresh when token expires. If it's 1min, then we will refresh a min before it expires... Etc.

Then we can get rid of the two knobs...

clients/go/admin/token_source_provider.go
```diff
@@ -141,15 +147,68 @@ func NewClientCredentialsTokenSourceProvider(ctx context.Context, cfg *Config,
...
+	}
 	return p.ccConfig.TokenSource(ctx), nil
 }
+
+type customTokenSource struct {
+	new                          oauth2.TokenSource
+	mu                           sync.Mutex // guards everything else
+	t                            *oauth2.Token
+	refreshTime                  time.Time
+	refreshed                    bool
```
EngHabu, line 178: This is really `failedToRefresh`. Right?

mayitbeegh: Yes, let me rename it.

EngHabu (approved):

katrogan (approved):

PR #287, opened 2022-04-21
Add interruptible override to execution #minor

MorpheusXAUT:
# TL;DR
Adds an `interruptible` flag to the `ExecutionSpec`, allowing for workflows to be flagged as interruptible for a single execution.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [ ] Unit tests added

## Complete description
An additional field has been added to the `ExecutionSpec`, allowing for the `interruptible` flag of a workflow to be override for a single execution (existing overrides on task level still remain in place). Default value of `false` falls back to workflow/task config.
The override is internally passed along via the `WorkflowExecutionSpec` in flyteadmin, some interfaces require the `LaunchPlanSpec` to implement the `interruptible` flag as well, although it is currently not used.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2284

## Follow-up issue
_NA_


protos/flyteidl/admin/execution.proto
```diff
@@ -289,6 +289,10 @@ message ExecutionSpec {
 
     // Controls how to select an available cluster on which this execution should run.
     ClusterAssignment cluster_assignment = 20;
+
+    // Allows for a workflow to be flagged as interruptible for a single execution.
+    // Omitting this field uses the workflow's defined value as a default.
+    bool interruptible = 21;
```
katrogan, line 295: since proto3 doesn't have a concept of required fields, the default value for bools will always be false. things get weird when the workflow sets interruptible to true but we explicitly want to disable interruptible for a single execution

can we use one-ofs here like we do with notifications to make setting this to false explicit (and not default): https://github.com/flyteorg/flyteidl/blob/master/protos/flyteidl/admin/execution.proto#L256

MorpheusXAUT: That's something I considered - the current implementation does not allow for an override to _disable_ interruptible for a single execution. I initially tried specing it with this in mind, however I believe using `oneof` does not work here due to the [`WorkflowExecutionConfigInterface`](https://github.com/flyteorg/flyteadmin/pull/410/files#diff-9b311b663f43eccf79220bfcae3a565ad26d83532c945f39cfc1f7ba029c22e3R451) used in flyteadmin - I did not find a way to get the generated `oneof` methods to play along nicely with the interface.

I did however attempt using the predefined [`google.protobuf.BoolValue`](https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/bool-value) in my [initial commit](https://github.com/flyteorg/flyteidl/pull/287/commits/c6be9b432163984bce2594d3f4925aee1a9f69c4#diff-724a7e8bb673471da21d8cb4a8ef634198bc10cc7b61965a920673c2b7408a54), however switched to using regular bools for simplicity. Using the `BoolValue` wrapper did produce compatible code though, so that could be used.
Alternatively, I could extract the interruptible override from flyteadmin's `mergeIntoExecConfig` so this incompatibility would not arise.

From our internal teams, I've heard that explicity _disabling_ was not a requirement, as they'd rather just enable it, but if that's something you'd be interested in, I can take another look.

katrogan: I think with the `WorkflowExecutionConfigInterface` you can change the method return type accordingly to the generated go struct for the oneof or alternate field instead, right? and determine the execution workflowExecConfig.Interruptible bit accordingly

in any case, cool! I didn't know about BoolValue but that seems like a nillable option we can use instead

also, sorry I'm not quite following on what you mean by

> Alternatively, I could extract the interruptible override from flyteadmin's mergeIntoExecConfig so this incompatibility would not arise and oneof could be used - I believe that would also get rid of the currently unused interruptible flag in the LaunchPlanSpec.

I think for the purpose of completeness and since this is an open source feature it would be nice to have a means of disabling as well

MorpheusXAUT: > I think with the `WorkflowExecutionConfigInterface` you can change the method return type accordingly to the generated go struct for the oneof or alternate field instead, right? and determine the execution workflowExecConfig.Interruptible bit accordingly

Yeah, but unfortunately the methods generated for the `oneof` of the `ExecutionSpec` and `LaunchPlanSpec` are not equal and not assignable to one another, at least from what I've seen. In any case, `BoolValue` works for all cases and solves the issue, so I'll go with that 🙂 

> also, sorry I'm not quite following on what you mean by
> 
> > Alternatively, I could extract the interruptible override from flyteadmin's mergeIntoExecConfig so this incompatibility would not arise and oneof could be used - I believe that would also get rid of the currently unused interruptible flag in the LaunchPlanSpec.

Instead of setting the interruptible flag in the `mergeIntoExecConfig` function, I could've done it at a different place, which wouldn't require the `WorkflowExecutionConfigInterface` to have an `GetInterruptible` method, but for the sake of consistency (and since `BoolValue` solves the issue anyways), I'll keep it there.

> I think for the purpose of completeness and since this is an open source feature it would be nice to have a means of disabling as well

I figured as much, changes are nearly ready and will be pushed tomorrow morning.

MorpheusXAUT: Changes using `BoolValue` have been pushed

protos/flyteidl/admin/execution.proto
```diff
@@ -289,6 +290,13 @@ message ExecutionSpec {
 
     // Controls how to select an available cluster on which this execution should run.
     ClusterAssignment cluster_assignment = 20;
+
+    // Allows for the interruptible flag of a workflow to be overwritten for a single execution.
+    // Omitting this field uses the workflow's value as a default.
```
katrogan, line 295:
```suggestion
    // Omitting this field uses the launch plan's value as a default.
```

protos/flyteidl/admin/execution.proto
```diff
@@ -289,6 +290,13 @@ message ExecutionSpec {
...
+    // As we need to distinguish between the field not being provided and its default value "false", we have to use a wrapper
+    // around the bool field. Since flyteadmin defines an interface requiring a unified interruptble override in several places, we cannot use
```
katrogan, line 297: can we omit the "Since flyteadmin defines an interface..." bit? I don't think we necessarily to dive into implementation details here. Also re: incompatible go types, couldn't we just have IsInterruptible return that generated oneof struct instead of a bool?

MorpheusXAUT: Sure thing, will omit the extra comment 👍 

Re: incompatible go types - I'm not 100% sure we could wrap it again since the `GetInterruptible()` method used is auto-generated for the [`ExecutionSpec`](https://github.com/MorpheusXAUT/flyteidl/blob/execution-interruptible/gen/pb-go/flyteidl/admin/execution.pb.go#L1209) and [`LaunchPlanSpec`](https://github.com/flyteorg/flyteidl/blob/c412ba26cf64ed81d95b5aa786a2d477b83c0852/gen/pb-go/flyteidl/admin/launch_plan.pb.go#L467). As those are used directly, we'd have to wrap them again somehow to make them usable with `oneof`.
If you'd prefer that instead of using the `BoolValue` wrapper, which works for both cases, I can have another look tomorrow.

protos/flyteidl/admin/launch_plan.proto
```diff
@@ -121,10 +121,9 @@ message LaunchPlanSpec {
     int32 max_parallelism = 18;
 
     // Allows for the interruptible flag of a workflow to be overwritten for a single execution.
-    // Omitting this field uses the workflow's value as a default.
+    // Omitting this field uses the launch plan's value as a default.
```
katrogan, line 124: oh wait i think this change doesn't make sense, let's keep the reference to workflow :)

katrogan (commented):
looks good, just one nit with comments!

MorpheusXAUT:
Not quite sure why the generate test fails, `DELTA_CHECK=true make generate` shows everything being fine locally for me :thinking: 
@katrogan any ideas? looks like some ordering of the generated documentation changed locally on my machine... Re-running `make generate` doesn't change anything for me though.

EngHabu:
This could be due to difference in tooling versions... Make sure you're using the same versions used in CI... Apologies about that

MorpheusXAUT:
@EngHabu Just re-generated the same on my Windows instead of Linux machine, seems to be working now... Not sure what exactly changed, as I believe I've installed the identical versions in both environments, but I'll have another look at that tomorrow.
Should go good now though 🙂

katrogan (approved):

katrogan:
@EngHabu mind reviewing too?

hamersaw (approved):

MorpheusXAUT:
@katrogan @hamersaw Any way to get this merged so I can update the other outstanding PRs with the updated version? :slightly_smiling_face:

PR #295, opened 2022-05-03
Dockerized docs gen

MorpheusXAUT:
# TL;DR
Docs generation currently still relies on locally installed versions of `protoc`, has now been changed to use Docker containers as well.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
The current setup for auto-generating documentation (within `generate_protos.sh`) still relies on a local version of `protoc` being installed, which is not managed/set up by boilerplate or repo dependencies either, requiring a new dev to set up the (correct) version locally themselves.
In order to allow for easier (and more consistent) docs generation, these calls have been changed to use the official `pseudomuto/protoc-gen-doc` as well, no longer requiring any local dependencies for the `generate_protos.sh` script.

The image used includes `protoc-gen-doc` version 1.4.1, which the [flyte fork](github.com/flyteorg/protoc-gen-doc) is built on. Whilst the flyte fork adds "native" support for reStructuredText templates, hardly any of the documentation generated uses the new type/template - except `core`, all other packages are generated using a custom template within the `flyteidl` repository anyways. I have thus decided to simply move the template used for `core` to the repository and keep it with the other, already existing one.
The fork's `restructuredtext` renderer was using `protoc-gen-doc`'s `HTMLRenderer`, all other custom templates default to the `TextRenderer` though. I believe this change should be irrelevant for this documented generation though as both templates don't seem to be using HTML exclusive functionality anyways (except HTML comment blocks, which I have replaced with the equivalent rst ones).
One change that _is_ relevant though is the use of double quotes in proto field comments: the `HTMLRenderer` URL-escapes those quotes while the `TextRenderer` doesn't, leading to incorrectly generated docs with missing field tables. This can be circumvented by using backticks or single quotes though - all existing `"` in comments have been replaced with `'` in the core protos.
Furthermore, the `pseudomuto/protoc-gen-doc:1.4.1` Docker image unfortunately only uses `protoc` version 3.6.1 (instead of 3.20.1 as used in the current GH workflow). Whilst the code generated seems to be identical (and only affects docs anyways), I'm not sure if that's a dealbreaker for this PR. We could bump the used image up to `pseudomuto/protoc-gen-doc:1.5.0` (the last one seems to be broken somehow), however we'd still be stuck with protoc 3.6.1...

Last but not least, I've set a locale override to `C.UTF-8` in the `generate_protos.sh` script. This ensures that the sorting of the generated documentation is consistent between all systems (a different locale was the reason for the [ordering issues](https://github.com/flyteorg/flyteidl/pull/287#issuecomment-1111501543) I had during my last PR).

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/2437

## Follow-up issue
_NA_


generate_protos.sh
```diff
@@ -4,6 +4,10 @@ DIR=`pwd`
 rm -rf $DIR/gen
 LYFT_IMAGE="lyft/protocgenerator:8167e11d3b3439373c2f033080a4b550078884a2"
 SWAGGER_CLI_IMAGE="docker.io/lyft/swagger-codegen-cli:dc5ce6ec6d7d4d980fa882d6bd13a83cba3be3c3"
+PROTOC_GEN_DOC_IMAGE="pseudomuto/protoc-gen-doc:1.4.1"
```
katrogan, line 7: I know this is a kinda tedious ask, but could we check in the Dockerfile and maybe even add automation (can also be filed as a todo/issue that someone else on the team can take up) so that we can easily update this image if need be?

katrogan (commented):
this is awesome! thank you for implementing this and fixing the issues with quotes

generate_protos.sh
katrogan: oh I didn't realize this was the official image, nevermind - please disregard this comment :-)

katrogan (approved):

generate_protos.sh
MorpheusXAUT: Yeah, using the official image for now since I figured that'd be easiest. We _could_ set up a build using the [fork](https://github.com/flyteorg/protoc-gen-doc), but I'm not sure that's worth for what seems to be needed/used at the moment.

katrogan:
PTAL @cosmicBboy @EngHabu

MorpheusXAUT:
Anything else needed to get this merged?
As we're currently planning our next Flyte changes that might require IDL adaptions, it'd be great to be able to use this instead of having to run docs generation on a different machine :sweat_smile:

kumare3:
@MorpheusXAUT and gang, also we should start planning for using buf build at some point

PR #296, opened 2022-05-03
Added GateNode message

hamersaw:
# TL;DR
GateNodes are used to stop the execution of downstream nodes. Simple gate node conditions are manual signalling or sleeping for a duration.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
^^^

## Tracking Issue
https://github.com/flyteorg/flyte/issues/208

## Follow-up issue
_NA_


protos/flyteidl/service/signal.proto
```diff
@@ -0,0 +1,55 @@
...
+
+import "google/api/annotations.proto";
+import "flyteidl/admin/signal.proto";
+import "protoc-gen-swagger/options/annotations.proto";
+
+// SignalService defines an RPC Service that may create, update, and retrieve signal(s).
+service SignalService {
+  // Fetches or creates a :ref:`ref_flyteidl.admin.Signal`.
+  rpc GetOrCreateSignal (flyteidl.admin.SignalGetOrCreateRequest) returns (flyteidl.admin.Signal) {
+    // Purposfully left out an HTTP API for this RPC call. This is meant to idempotently retrieve
```
katrogan, line 14: hopefully one day we'll have authz!

protos/flyteidl/admin/signal.proto
```diff
@@ -0,0 +1,87 @@
...
+    // +required
+    uint32 limit = 2;
+
+    // In the case of multiple pages of results, the, server-provided token can be used to fetch the next page
+    // in a query.
+    // +optional
+    string token = 3;
+
+    // Indicates a list of filters passed as string.
+    // More info on constructing filters : <Link>
```
katrogan, line 38: nit: can we populate the `<Link>` with a sphinx reference

protos/flyteidl/admin/signal.proto
```diff
@@ -0,0 +1,87 @@
...
+    core.Literal value = 2;
+}
+
+// SignalSetResponse represents a response structure if signal setting succeeds.
+message SignalSetResponse {
+    // Purposefully empty, may be populated in the future.
+}
+
+// Signal encapsulates a unique identifier, associated metadata, and a value for a single Flyte
+// signal. Signals may exist either without a set value (representing a signal request) or with a
```
katrogan, line 75: interesting, can i ask why use an implicit behavior and not two separate messages for clarity instead?

protos/flyteidl/core/workflow.proto
```diff
@@ -129,6 +155,9 @@ message Node {
 
         // Information about the branch node to evaluate in this node.
         BranchNode branch_node = 8;
+
+        // TODO hamersaw - document
```
katrogan, line 159: +1

protos/flyteidl/core/workflow.proto
```diff
@@ -68,6 +68,32 @@ message WorkflowNode {
     }
 }
 
+// SignalCondition represents a dependency on an existing signal.
+message SignalCondition {
+    // A unique identifier for the requested signal.
+    string signal_id = 1;
+
+    // A type denoting the required value type for this signal.
+    LiteralType type = 2;
```
katrogan, line 77: Are there broader docs for how this is used? I didn't see an issue as part of the PR and I don't think I quite follow how this eventual value is produced. Could there be a case where we want multiple values as part of a signal?

hamersaw: The best resource is probably the [RFC](https://github.com/flyteorg/flyte/pull/2090). Currently the implementation is setup so FlytePropeller calls the GetOrCreate RPC endpoint while evaluating the GateNode during each iteration. The node only succeeds if the signal has a value set.

An external entity will set the signal value. For example, a user through the SetSignal RPC endpoint or the UI can set the signal with a post to the RESTful interface.

And I believe the LiteralType proto encapsulates a Collection, Map, etc. So we should be able to support a variety of variable types, including multiple values.

protos/flyteidl/core/workflow.proto
hamersaw: Thanks. Slipped past me!

protos/flyteidl/admin/signal.proto
hamersaw: I thought it may be easier for the model transformers within FlyteAdmin, write one and use if everywhere. I think using a single table in FlyteAdmin with both the type and value is the correct approach, so emulating that in proto messages. There does not seem to be strong uniformity between the different model types. I am open to changing this if anyone feels strongly.

protos/flyteidl/service/signal.proto
hamersaw: Does FlyteAdmin have a scheme implemented already? Happy to add something existing. I'm sure starting from the ground up would be a bit out of scope for this work.

protos/flyteidl/admin/signal.proto
hamersaw: To be honest, I just copied the list proto from other messages. It doesn't look like any have a link populated. Does one exist? Otherwise I will just remove the link.

protos/flyteidl/core/workflow.proto
```diff
@@ -68,6 +68,35 @@ message WorkflowNode {
     }
 }
 
+// SignalCondition represents a dependency on an existing signal.
```
wild-endeavor, line 71: Can you clarify what you mean by an existing signal?

hamersaw: I was probably caught up in the flyteadmin side; where a signal does not technically "exist", meaning it can be provided, until propeller processes the gatenode that contains the signal condition. I'll go ahead and remove this, it can be confusing.

wild-endeavor (approved):

wild-endeavor (approved):

protos/flyteidl/service/signal.proto
eapolinario, line 14: typo: Purposfully

eapolinario (approved):

wild-endeavor (approved):

EngHabu (approved):

PR #298, opened 2022-05-19
Add grpc load balancing configs

ckiosidis:
# TL;DR

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/2510

## Follow-up issue
_NA_
OR

### DIFF
--- a/clients/go/admin/client.go
+++ b/clients/go/admin/client.go
@@ -19,7 +19,7 @@ import (
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/health/grpc_health_v1"
 
-	_ "google.golang.org/grpc/balancer/grpclb" //nolint
+	_ "google.golang.org/grpc/balancer/grpclb"     //nolint
 	_ "google.golang.org/grpc/balancer/roundrobin" //nolint
 )
 
--- a/clients/go/admin/client.go
+++ b/clients/go/admin/client.go
@@ -19,8 +19,8 @@ import (
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/health/grpc_health_v1"
 
-	_ "google.golang.org/grpc/balancer/grpclb"
-	_ "google.golang.org/grpc/balancer/roundrobin"
+	_ "google.golang.org/grpc/balancer/grpclb" //nolint
+	_ "google.golang.org/grpc/balancer/roundrobin" //nolint
 )
 
 // IDE "Go Generate File". This will create a mocks/AdminServiceClient.go file
--- a/clients/go/admin/client.go
+++ b/clients/go/admin/client.go
@@ -192,6 +192,10 @@ func initializeClients(ctx context.Context, cfg *Config, tokenCache pkce.TokenCa
 		opts = append(opts, authOpt)
 	}
 
+	if cfg.Balancer != "" {
+		opts = append(opts, grpc.WithDefaultServiceConfig(fmt.Sprintf(`{"loadBalancingPolicy":"%v"}`, cfg.Balancer)))
+	}
+
 	adminConnection, err := NewAdminConnection(ctx, cfg, opts...)
 	if err != nil {
 		logger.Panicf(ctx, "failed to initialize Admin connection. Err: %s", err.Error())
--- a/clients/go/admin/client.go
+++ b/clients/go/admin/client.go
@@ -199,7 +199,6 @@ func initializeClients(ctx context.Context, cfg *Config, tokenCache pkce.TokenCa
 		// The grpcClient will be created regardless of the value of the loadBalancingPolicy.
 		// When a balancing policy cannot be found, the client will be created with `pick_first`.
 		// available policies can be found here https://github.com/grpc/grpc/blob/master/doc/load-balancing.md#load-balancing-policies
-		// TODO: test that the loadBalancingPolicy is configured properly and validate towards the value from the grpc.clientConn
 		opts = append(opts, grpc.WithDefaultServiceConfig(fmt.Sprintf(`{"loadBalancingPolicy":"%v"}`, cfg.LoadBalancingPolicy)))
 	}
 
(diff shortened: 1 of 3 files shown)


clients/go/admin/client.go
```diff
@@ -192,6 +192,11 @@ func initializeClients(ctx context.Context, cfg *Config, tokenCache pkce.TokenCa
 		opts = append(opts, authOpt)
 	}
 
+	if cfg.Balancer != "" {
```
honnix, line 195: nit: maybe `BalancerName`?

clients/go/admin/client_test.go
```diff
@@ -276,3 +276,33 @@ func ExampleClientSetBuilder() {
 	_ = clientSet.AuthMetadataClient()
 	_ = clientSet.IdentityClient()
 }
+
+func TestBalancerConfigs(t *testing.T) {
```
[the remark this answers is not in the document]
ckiosidis: it won't throw errors unfortunately. If the balancer cannot be found it will default to pick_first

clients/go/admin/client.go
ckiosidis: changed it to LoadBalancingPolicy, what do you think?

clients/go/admin/client_test.go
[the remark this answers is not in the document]
honnix: Maybe we should preload `roundrobin` package. Maybe `grpclb` as well but that seems to be deprecated.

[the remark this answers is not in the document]
ckiosidis: it didnt seem to be an issue in the tests, but i can add it in the next commit

honnix:
> Update the comment as well? https://github.com/flyteorg/flyteidl/blob/master/clients/go/admin/config.go#L1-L4

This is not yet addressed.

clients/go/admin/client_test.go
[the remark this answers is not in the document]
honnix: I meant something like `import _ "google.golang.org/grpc/balancer/roundrobin"` in `client.go`.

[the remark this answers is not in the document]
ckiosidis: yeah i realised that i added it in the wrong file

clients/go/admin/client.go
```diff
@@ -192,6 +195,13 @@ func initializeClients(ctx context.Context, cfg *Config, tokenCache pkce.TokenCa
 		opts = append(opts, authOpt)
 	}
 
+	if cfg.LoadBalancingPolicy != "" {
+		// The grpcClient will be created regardless of the value of the loadBalancingPolicy.
```
honnix, line 199: Let's not to repeat the comment?

honnix (approved):

honnix:
Let's not try to close https://github.com/flyteorg/flyte/issues/2510 because datacatalog is not yet addressed. @ckiosidis

clients/go/admin/client.go
```diff
@@ -192,6 +195,10 @@ func initializeClients(ctx context.Context, cfg *Config, tokenCache pkce.TokenCa
 		opts = append(opts, authOpt)
 	}
 
+	if cfg.LoadBalancingPolicy != "" {
+		opts = append(opts, grpc.WithDefaultServiceConfig(fmt.Sprintf(`{"loadBalancingPolicy":"%v"}`, cfg.LoadBalancingPolicy)))
```
EngHabu, line 199: Should we expose the ServiceConfig as the config? instead of the load balancing policy?
Docs are here: https://github.com/grpc/grpc/blob/master/doc/service_config.md
Protobuf type is here: https://github.com/grpc/grpc-proto/blob/master/grpc/service_config/service_config.proto#L625

It looks like there is no current Go type (the existing one is still used by the code but marked as deprecated)... so we can make this config a string with links to the documentation.

wdyt @ckiosidis, @honnix ?

honnix: Right. That makes sense. Is there any other package we will need to preload?

ckiosidis: I pushed a change, let me know what you think @EngHabu

clients/go/admin/config.go
```diff
@@ -68,6 +67,11 @@ type Config struct {
 	PkceConfig pkce.Config `json:"pkceConfig" pflag:",Config for Pkce authentication flow."`
 
 	Command []string `json:"command" pflag:",Command for external authentication token generation"`
+
+	// Set the ServiceConfig formatted as a json string https://github.com/grpc/grpc/blob/master/doc/service_config.md
+	// eg. { "loadBalancingConfig": [{ "round_robin":{} }], "methodConfig": [{ "name": [{ "service": "foo", "method": "bar" }, { "service": "baz" }], "timeout": "1.000000001s"}]}
+	// find the full schema here https://github.com/grpc/grpc-proto/blob/master/grpc/service_config/service_config.proto#L625
+	ServiceConfig string `json:"serviceConfig" pdflag:",Set the service config for the grpc client"`
```
honnix, line 74: Maybe `DefaultServiceConfig` reflects it better because it is the default if a service doesn't enforce override. ;)

clients/go/admin/client.go
```diff
@@ -18,6 +18,9 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/health/grpc_health_v1"
+
+	_ "google.golang.org/grpc/balancer/grpclb"     //nolint
```
honnix, line 22: I think this is deprecated.

```suggestion
```

clients/go/admin/client.go
```diff
@@ -192,6 +195,10 @@ func initializeClients(ctx context.Context, cfg *Config, tokenCache pkce.TokenCa
 		opts = append(opts, authOpt)
 	}
 
+	if cfg.ServiceConfig != "" {
+		opts = append(opts, grpc.WithDefaultServiceConfig(cfg.ServiceConfig))
```
honnix, line 199:
```suggestion
		opts = append(opts, grpc.WithDefaultServiceConfig(cfg.DefaultServiceConfig))
```

ckiosidis:
fixed on https://github.com/flyteorg/flyteidl/pull/301

PR #315, opened 2022-09-02
Use grpc client interceptors to properly check for auth requirement

EngHabu:
Signed-off-by: Haytham Abuelfutuh <haytham@afutuh.com>

# TL;DR
Change client auth mechanism to attempt to build a TokenSource whenever any client call received the equivalent of http 401 UNAUTHORIZED response code.

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Existing mechanism to build an authenticated client depends on auth negotiation at the very first time the gRPC client connection is built. If at the time:
1. Auth hasn't been enabled on admin, or
2. It fails to query for Auth endpoints for any transient network reasons.
It'll fallback to assuming Auth isn't enabled on the server and build an unauthenticated gRPC client. Once that's built, it's reused for the entirety of the lifetime of the process.
If auth is enabled (or config changed) at a later point, all existing clients will fail to authenticate and never recover.

With this change, it establishes a better pattern akin to the HTTP-recognized behavior of responding to auth code failures by attempting to authenticate.

After the change, the client will always be built unauthenticated, as soon as the first request is sent and receives an authentication error, it'll attempt to negotiate auth and build a valid token source.

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/2841

### DIFF
--- a/clients/go/admin/auth_interceptor.go
+++ b/clients/go/admin/auth_interceptor.go
@@ -19,8 +19,6 @@ import (
 func MaterializeCredentials(ctx context.Context, cfg *Config, tokenCache cache.TokenCache, perRPCCredentials *PerRPCCredentialsFuture) error {
 	authMetadataClient, err := InitializeAuthMetadataClient(ctx, cfg)
 	if err != nil {
-		// EngHabu: I don't know why did we use to panic here. Generally speaking, libraries should error on the side of
-		// 	returning errors instead.
 		return fmt.Errorf("failed to initialized Auth Metadata Client. Error: %w", err)
 	}
 
--- a/clients/go/admin/auth_interceptor.go
+++ b/clients/go/admin/auth_interceptor.go
@@ -59,17 +59,20 @@ func shouldAttemptToAuthenticate(errorCode codes.Code) bool {
 func newAuthInterceptor(cfg *Config, tokenCache cache.TokenCache, credentialsFuture *PerRPCCredentialsFuture) grpc.UnaryClientInterceptor {
 	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
 		err := invoker(ctx, method, req, reply, cc, opts...)
-		logger.Debugf(ctx, "Request failed due to [%v]. If it's an unauthenticated error, we will attempt to establish an authenticated context.", err)
-		if st, ok := status.FromError(err); ok {
-			// If the error we receive from executing the request expects
-			if shouldAttemptToAuthenticate(st.Code()) {
-				logger.Debugf(ctx, "Request failed due to [%v]. Attempting to establish an authenticated connection and trying again.", st.Code())
-				newErr := MaterializeCredentials(ctx, cfg, tokenCache, credentialsFuture)
-				if newErr != nil {
-					return fmt.Errorf("authentication error! Original Error: %v, Auth Error: %w", err, newErr)
-				}
+		if err != nil {
+			logger.Debugf(ctx, "Request failed due to [%v]. If it's an unauthenticated error, we will attempt to establish an authenticated context.", err)
 
-				return invoker(ctx, method, req, reply, cc, opts...)
+			if st, ok := status.FromError(err); ok {
+				// If the error we receive from executing the request expects
+				if shouldAttemptToAuthenticate(st.Code()) {
+					logger.Debugf(ctx, "Request failed due to [%v]. Attempting to establish an authenticated connection and trying again.", st.Code())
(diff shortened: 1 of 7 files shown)


EngHabu:
@mayitbeegh @pmahindrakar-oss @katrogan still WIP but please take an initial look at how this is refactored... and provide any early feedback you got. Thank you!

clients/go/admin/auth_interceptor.go
```diff
@@ -0,0 +1,74 @@
...
+// It will first invoke the grpc pipeline (to proceed with the request) with no modifications. It's expected for the grpc
+// pipeline to already have a grpc.WithPerRPCCredentials() DialOption. If the tokenSource has already been initialized,
+// it'll take care of refreshing when it expires... etc.
+// If the first invocation fails with an auth error, this interceptor will then attempt to establish a token source once
+// more. It'll fail hard if it couldn't do so (i.e. it will no longer attempt to send an unauthenticated request). Once
+// a token source has been created, it'll invoke the grpc pipeline again, this time the grpc.PerRPCCredentials should
+// be able to find and acquire a valid AccessToken to annotate the request with.
+func newAuthInterceptor(cfg *Config, tokenCache cache.TokenCache, tokenSource *CustomHeaderTokenSource) grpc.UnaryClientInterceptor {
+	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+		err := invoker(ctx, method, req, reply, cc, opts...)
```
pmahindrakar-oss, line 58: Can we may be log this err too

clients/go/admin/auth_interceptor.go
```diff
@@ -0,0 +1,74 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to fetch client metadata. Error: %v", err)
+	}
+
+	tSource, err := tokenSourceProvider.GetTokenSource(ctx)
+	if err != nil {
+		return err
+	}
+
+	tokenSource.Reset(tSource, clientMetadata.AuthorizationMetadataKey, cfg.UseInsecureConnection)
```
pmahindrakar-oss, line 42: we are using tokenSource and tSource which is confusing to read. May be can tokenSource be renamed to singletonTokenSource

pmahindrakar-oss (commented):
LGTM

clients/go/admin/auth_interceptor.go
```diff
@@ -0,0 +1,74 @@
...
+	return nil
+}
+
+// newAuthInterceptor creates a new grpc.UnaryClientInterceptor that forwards the grpc call and inspects the error.
+// If an auth error is detected, it'll panic the process. This is useful for scenarios when auth is enabled after the client
```
katrogan, line 47: does this actually panic? do we want it to instead of retrying with a rebuilt client?

clients/go/admin/auth_interceptor.go
EngHabu: plz let me know what do you think now

clients/go/admin/auth_interceptor.go
EngHabu: oh good catch! I guess I was doing some next level DDD (Documentation Driven Development. For a split sec I thought I just coined that but google disappointed me!).

So it does try with the rebuilt client... I initially thought to make it panic to force a restart... but I don't think we need to anymore because it will attempt to rebuild the client on every subsequent call until it manages to get through... so I'll update the docs... plz let me know what you think.

clients/go/admin/auth_interceptor_test.go
```diff
@@ -0,0 +1,205 @@
...
+		f := NewPerRPCCredentialsFuture()
+		interceptor := newAuthInterceptor(&Config{}, &mocks.TokenCache{}, f)
+		otherError := func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, opts ...grpc.CallOption) error {
+			return status.New(codes.Canceled, "").Err()
+		}
+
+		assert.Error(t, interceptor(context.Background(), "POST", nil, nil, nil, otherError))
+	})
+
+	t.Run("Unauthenticated first time, succeed", func(t *testing.T) {
```
katrogan, line 128: is it possible to add a test for the authenticated code path or is that too hairy for unit tests?

katrogan (approved):

katrogan (dismissed):

pmahindrakar-oss (approved):

katrogan (approved):

PR #329, opened 2022-10-12
Adding configurable audience property for flyte clients

pmahindrakar-oss:
Signed-off-by: pmahindrakar-oss <prafulla.mahindrakar@gmail.com>

# TL;DR
Most Authorization server provide a way to use default audience which can be used to request auth tokens which is what flyte relied on before this PR.

The current PR allows to send a configurable audience field when requesting auth token from the authorization server.

Follow up PR to return audience also from PublicClientConfig endpoint of flyteadmin

https://github.com/flyteorg/flyteadmin/pull/485

If client config contains the audience field then that would be used 
Though if the override to useAudienceFromAdmin is set then whatever audience is set in admin would be used

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2959

## Follow-up issue
_NA_

### DIFF
--- a/clients/go/admin/token_source_provider.go
+++ b/clients/go/admin/token_source_provider.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"strings"
 	"sync"
@@ -20,6 +21,10 @@ import (
 	"github.com/flyteorg/flyteidl/clients/go/admin/tokenorchestrator"
 	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/service"
 	"github.com/flyteorg/flytestdlib/logger"
+)
+
+const (
+	audience = "audience"
 )
 
 // TokenSourceProvider defines the interface needed to provide a TokenSource that is used to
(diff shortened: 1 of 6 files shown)


clients/go/admin/token_source_provider.go
```diff
@@ -45,16 +50,21 @@ func NewTokenSourceProvider(ctx context.Context, cfg *Config, tokenCache cache.T
 			tokenURL = metadata.TokenEndpoint
 		}
 
+		clientMetadata, err := authClient.GetPublicClientConfig(ctx, &service.PublicClientAuthConfigRequest{})
```
katrogan, line 53: should we only fetch this if the `cfg.Audience` is empty?

pmahindrakar-oss: Done., if either scopes or Audience is empty then we call this api

clients/go/admin/token_source_provider.go
```diff
@@ -46,15 +51,24 @@ func NewTokenSourceProvider(ctx context.Context, cfg *Config, tokenCache cache.T
 		}
 
 		scopes := cfg.Scopes
-		if len(scopes) == 0 {
-			clientMetadata, err := authClient.GetPublicClientConfig(ctx, &service.PublicClientAuthConfigRequest{})
+		audienceValue := cfg.Audience
+
+		if len(scopes) == 0 || len(audienceValue) == 0 {
```
katrogan, line 56: If I'm understanding correctly with this change now we'll always make the call to get the public client config even in the case where the audience property is not actually required? (It isn't always required, right - this is just custom for some deployments?)

Is there anyway to indicate we want to explicitly fetch the audience property so that end users don't have to now update their deployments with a dummy config value for the audience property to avoid this lookup even when scopes are set?

pmahindrakar-oss: Done Added added new config flag `useAudienceFromAdmin` which will force pull config from Admin only if its set otherwise it wont call the publicEndpoint config

katrogan (approved):

wild-endeavor (approved):

PR #338, opened 2022-10-31
Docker dev container

daniel-shuy:
# TL;DR
Add Docker dev container to have consistent, reproducible dev environments

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

The goal is to run `make` targets in a Docker container to have a consistent, reproducible dev environment (same version of Golang and tooling).

Initially I was thinking of using the approach detailed in https://vbehar.medium.com/makefile-tip-of-the-day-run-with-or-without-docker-9c00ad84a700, e.g.
```makefile
# Makefile

ifneq ($(SKIP_DOCKER),true)
	PROJECT_ROOT := $(dir $(realpath $(firstword $(MAKEFILE_LIST))))

	DOCKER_IMG := golang:1.18.5
	DOCKER_CMD := docker run \
		-it \
		--rm \
		-e SKIP_DOCKER=true \
		-e REPOSITORY=$(REPOSITORY) \
		-v $(PROJECT_ROOT):/usr/src/flyteidl \
		-w /usr/src/flyteidl \
		$(DOCKER_IMG)
endif
```

Then it can be added before any command to run it in a Docker container, or locally if `SKIP_DOCKER` is set to `true`, e.g.
```makefile
.PHONY: generate
generate: update_boilerplate install doc_gen_deps # get latest boiler plate, install tools, generate protos, mock, pflags and  get doc dependencies
	./generate_protos.sh
	$(DOCKER_CMD) ./generate_mocks.sh
	$(DOCKER_CMD) go generate ./...
```

Unfortunately this doesn't work because `generate` depends on `install`, which is from `boilerplate/flyte/golang_test_targets/Makefile`. To use this approach will require modifying `boilerplate/flyte/golang_test_targets/Makefile` to run in the container as well, but `boilerplate/flyte/golang_test_targets/Makefile` is from another repository (https://github.com/flyteorg/boilerplate).

Therefore I went with my current approach, using a Dockerfile to build a Docker image with tools and dependencies installed (using `make install`), and using a Docker Compose file to map the volume.

And while I was at it, I also decided to add support for VS Code's Dev Containers (https://code.visualstudio.com/docs/devcontainers/create-dev-container#_extend-your-docker-compose-file-for-development).

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/2911

## Follow-up issue
_NA_


devcontainer.Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM golang:1.18.5
```
daniel-shuy, line 1: Is `1.18.5` the correct Golang version?

eapolinario (approved):

devcontainer.Dockerfile
eapolinario: This is ok. Most people are using a flavor of 1.18 anyways. 

We're also investing in a monorepo, which will make this kind of standardization a lot easier.

hamersaw:
@daniel-shuy I tried to run this locally and I think there may be a bit more tweaking to do. So after starting the compose container and entering it, the call to `make generate` is failing because the `generate-protos.sh` script requires docker (which is not in our dev container). Of course this is still useful for calling `generate-mocks.sh` and `go generate ./...`, but I'm wondering if modifying the Makefile is the correct way to clean this up in the near-term.

daniel-shuy:
@hamersaw ah, right. Maybe we should use [docker:dind](https://hub.docker.com/_/docker) as the base image and install `golang` in it instead

hamersaw:
Going to punt on this until we get the monorepo setup and we can create a singular build setup for all Flyte components.

PR #346, opened 2022-11-28
Use buf to generate python stubs

eapolinario:
# TL;DR
Use buf to generate python stubs and pyi files

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
As described in https://developers.google.com/protocol-buffers/docs/news/2022-05-06, starting on version 4.21.0, the python protobuf package is based on the upb library. This requires us to remove the upper bound we put on the `protobuf` package if we are to leverage these new features. We are also receiving a fair amount of questions regarding the supported version of `protobuf` from Flyte users, so all in all it's a good time to relax that constraint.

This PR does two things:
1. Use buf to generate the python stubs
2. Generate the pyi files, which unlocks the removal of model files in flytekit.

## Tracking Issue

## Follow-up issue
_NA_
OR


wild-endeavor (approved):

.github/workflows/verification.yml
```diff
@@ -51,19 +51,20 @@ jobs:
...
-        run: python3 /usr/local/bin/entrypoint.py -i ./protos -d protos/flyteidl/admin -l python
-      - name: Proto-Core-Python
-        run: python3 /usr/local/bin/entrypoint.py -i ./protos -d protos/flyteidl/core -l python
-      - name: Proto-Event-Python
-        run: python3 /usr/local/bin/entrypoint.py -i ./protos -d protos/flyteidl/event -l python
-      - name: Proto-Plugins-Python
-        run: python3 /usr/local/bin/entrypoint.py -i ./protos -d protos/flyteidl/plugins -l python
-      - name: Proto-Datacatalog-Python
-        run: python3 /usr/local/bin/entrypoint.py -i ./protos -d protos/flyteidl/datacatalog -l python
+      # - name: Proto-Service-Python
```
pingsutw, line 56: should we remove it?

pingsutw (approved):

wild-endeavor (approved):

.github/workflows/verification.yml
eapolinario: This will all be removed once we move the other plugins to buf. Leaving it like this is a good indication that we're in this in-between state.

PR #351, opened 2022-12-07
Add Databricks config to Spark config

pingsutw:
# TL;DR
This config is used by the dbx plugin in propeller to submit the dbx job request.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
^^^

## Tracking Issue
_NA_

## Follow-up issue
_NA_


protos/flyteidl/plugins/spark.proto
```diff
@@ -21,4 +21,8 @@ message SparkJob {
     map<string, string> sparkConf = 4;
     map<string, string> hadoopConf = 5;
     string executorPath = 6; // Executor path for Python jobs.
+    // databricksConf is base64 encoded string which stores databricks job configuration.
```
wild-endeavor, line 24: why is the databricks one a b64 encoded string while the spark and hadoop conf's above just a map of str, str?  Is the databricks one a binary object that needs to be encoded?

wild-endeavor (approved):
approve but maybe address question in a comment?

protos/flyteidl/plugins/spark.proto
pingsutw: Because it's nested dict. wait, I think I can use `Struct` here. Let me update it

tmathur-hbo: +1

wild-endeavor (approved):

eapolinario (approved):

PR #358, opened 2023-01-11
Added template configuration to task template

hamersaw:
# TL;DR
This PR adds a template field to the `TaskTemplate` which contains either a `PodTemplate` name or a marshaled `PodTemplate` struct.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
^^^

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3123

## Follow-up issue
_NA_


protos/flyteidl/core/tasks.proto
```diff
@@ -143,6 +143,15 @@ message TaskTemplate {
     // to use as required.
     // reserve the field numbers 1 through 15 for very frequently occurring message elements
     map<string, string> config = 16;
+
+    // pod_template provides the definition for a PodTemplate to be applied as the base configuration for this task.
+    // This enables fine-grained k8s configuration for Flyte tasks in addition to overriding the default PodTemplate.
+    oneof pod_template {
+        // pod_template_name is the unique name of a PodTemplate k8s resource.
+        string pod_template_name = 19;
+        // pod_template_struct is a JSON marshaled PodTemplate.
```
EngHabu, line 152: Can you elaborate here on how overriding properties will work? Will service account be copied from the task instead.. how about container images... Etc. Etc.

EngHabu (approved):
Let's add more code comments to explain the behavior. I'm pretty sure we will get a lot of questions here...

protos/flyteidl/core/tasks.proto
hamersaw: So we have put together pretty [extensive docs on configuring Flyte pods](https://docs.flyte.org/en/latest/deployment/cluster_config/general.html#configuring-k8s-pod) which highlights using the default PodTemplate vs k8s plugin configuration. I imagined we might extend on this (maybe make a separate page?) to try and sort this out end to end. How in-depth do you think is necessary here? We could fill 100 lines I imagine.

kumare3:
Why do we need this, why not use the one of

flixr:
@kumare3 e.g. for my usecases it boils down to setting some things like `runtimeClassName`, `affinity`, etc. per task (some tasks in the workflow need to run on nodes with large GPU, same only need a small one). And it would allow mounting of secrets, etc.. see also https://github.com/flyteorg/flyte/issues/3055

eapolinario (approved):

EngHabu (approved):

PR #364, opened 2023-02-10
add partition_columns to StructuredDatasetType

cosmicBboy:
Signed-off-by: Niels Bantilan <niels.bantilan@gmail.com>

## Add `partition_columns` to `StructuredDatasetType`

Partially addresses https://github.com/flyteorg/flyte/issues/3219

# TL;DR

This PR adds an additional property to the `StructureDatasetType` protobuf definition so that metadata about which columns in the dataset (some kind of DataFrame object) are used for partitioning the dataset into chunks, for example when a `pandas.DataFrame` is serialized as a parquet file.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

This change is required to store additional metadata about which columns are used for partitioning. Currently this only meaningfully affects the serialization/deserialization of parquet files, but in the future we could support the partitioning of other serialization formats.

## Tracking Issue

Partly addresses https://github.com/flyteorg/flyte/issues/3219

## Follow-up issue
NA


protos/flyteidl/core/types.proto
```diff
@@ -67,6 +67,11 @@ message StructuredDatasetType {
     // The serialized bytes of a third-party schema library like Arrow.
     // This is an optional field that will not be used for type checking.
     bytes external_schema_bytes = 4;
+    
+    // This indicates which columns in the structured dataset is used for
+    // partitioning when the dataset is serialized. This is an optional field
+    // that will not be userd for type checking
```
hamersaw, line 73: datset is used -> dataset *are* used

userd -> used

period on the end?

hamersaw (commented):
@eapolinario please confirm, did we decide this needed to be included in the serialized flyteidl type or could just be read from metadata within flytekit at runtime?

protos/flyteidl/core/types.proto
cosmicBboy: Nice catch! I think dataset is singular :)

cosmicBboy:
> @eapolinario please confirm, did we decide this needed to be included in the serialized flyteidl type or could just be read from metadata within flytekit at runtime?

I may be missing something, but we need to include it in the type so that the [structured dataset decoder](https://github.com/flyteorg/flytekit/blob/47ac6ac2a547fdd2b46654db5163493b4f33dbb2/flytekit/types/structured/basic_dfs.py#L47-L63) has access to the metadata (unless we want to manually inspect the uri path for multiple directories)

protos/flyteidl/core/types.proto
hamersaw: but columns is plural right?

cosmicBboy: 🤦‍♂️

PR #381, opened 2023-03-09
Init customTokenSource.refreshTime

andrewwdye:
# TL;DR
Follow up to https://github.com/flyteorg/flyteidl/pull/377. This change sets `refreshTime` based on the cached token expiry when instantiating `customTokenSource`. Otherwise `refreshTime` defaults to 0, effectively nullifying the cache we just added since it's always viewed as refresh window exceeded on first use.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
This change
* Chooses a new `refreshTime` if a cached token is available, otherwise default to empty time
* Moves token source provider tests to dedicated file
* Adds new test `TestCustomTokenSource_GetTokenSource`
* Removes overly verbose logging

## Tracking Issue
NA

## Follow-up issue
NA


eapolinario (approved):
good catch!

PR #386, opened 2023-03-31
Refactor kf-operator plugins configs and support setting different specs for different replica groups

yubofredwang:
# TL;DR
This PR refactors config fields for kubeflow operator plugins including: TFJob, MPIJob, PyTorchJob. This change is **backward incompatible**, so we will keep the original plugin configs untouched and use the new configs in v2 version of the plugins.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Code change includes:
1. Move TFJob, MPIJob, PyTorchJob from `flyteidl/plugins/` to `flyteidl/plugins/kubeflow/`
2. Create `flyteidl/plugins/kubeflow/common.proto` that contains the definition RestartPolicy, RunPolicy and CleanPodPolicy
3. Add `xxxReplicaSpec` in each job type to allow settings of `replicas`, `image`, `resources` and `restart_policy`

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/3308


hamersaw:
Quick update per our offline discussion. The goal is to make these changes backwards compatible with the existing flytekit kubeflow operator APIs. So the proposed IDL spec will override the existing approach, so the kubeflow operator updates are something like:
```
@task(
    task_config=TfJob(
        num_workers=2, # @deprecated
        num_ps_replicas=1, # @deprecated
        num_chief_replicas=1 # @deprecated
        worker_spec=TfWorkflowSpec(
             replicas=3 # backend job will use 3 replicas, overriding the 2 provided above
        ),
        # similar specs for ps and chief replicas
    ),
)
def foo:
    # omitted
    
@task(
    task_config=TfJob(
        num_workers=2, # @deprecated - but used because worker_spec is not specified
        num_ps_replicas=1, # @deprecated - similarly used
        num_chief_replicas=1 # @deprecated  -similarly used
    ),
)
def foo:
    # omitted
```
IIUC this is the cleanest way to implement these changes. Very interested in flytekit maintainer thoughts - cc @wild-endeavor @eapolinario.

The largest open question is how this should be written in wire-format, there are two solutions:
(1) the proposed solution creates a new flyteidl proto message to support the new format. This will require flytekit to compile all new kubeflow jobs to use `TaskType=1` (rather than the exist `0`) in the `TaskTemplate` denoting the `custom` struct is using the old or new configuration definition. In this approach flytekit will be responsible for applying the override values and the deprecation messages will be on the flytekit kubeflow API fields.
(2) Merge the new and existing proto messages. In this scenario the idl message will directly reflect the kubeflow APIs. In the above example, the idl message will have both `num_workers` and `workers_spec.replicas`. In this approach flyteplugins will be responsible for applying the override values and the deprecation messages will be on both the flytekit kubeflow API fields and the flyteidl configuration proto.

@yubofredwang please feel free to correct this and / or clarify if there is more context here.

hamersaw:
@yubofredwang we talked internally here, (hopefully) quick summary. The best approach is probably to use a new proto message and increment the TaskType version to 1. This means deprecating the top-level `replica` counts will be done in flytekit-side. Does this sound reasonable to you? We should be able to move forward with this PR then, but want to touch-base before pushing it through.

yubofredwang:
"The best approach is probably to use a new proto message"

so

yubofredwang:
> @yubofredwang we talked internally here, (hopefully) quick summary. The best approach is probably to use a new proto message and increment the TaskType version to 1. This means deprecating the top-level `replica` counts will be done in flytekit-side. Does this sound reasonable to you? We should be able to move forward with this PR then, but want to touch-base before pushing it through.

Got it. My understanding is we make the following changes:

flyteidl: keep the new proto separated from the old protos of kf-operators
flyteplugins: support both TaskTypeVersion 0 and 1, the TaskTemplate will be in different format for the two versions
flytekit: bump the flytekit version and use the new proto only in the new flytekit version

My questions regarding to this approach is how the user upgrade process should look like. There are three related pieces here: user code, flytekit, flyteplugins

Case1: code unchanged, flytekit unchanged, flyteplugins updated
in this case, the existing workflow works without a problem since flyteplugins support both version

Case2: code unchanged, flytekit updated, flyteplugins unchanged
this use case, the existing workflow will fail due to flytekit does not accept old config. User should get error message instructing them to update code.

Case3: code updated, flytekit updated, flyteplugins unchanged
same as case1

Case4: code updated, flytekit unchanged, flyteplugins updated
Not likely to occur since users are not aware of the new flytekit interface

ByronHsu:
Regardless of the compatibility issues. The idl looks good to me!

protos/flyteidl/plugins/kubeflow/pytorch.proto
```diff
@@ -7,6 +7,16 @@ option go_package = "github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/plugins";
 import "flyteidl/core/tasks.proto";
 import "flyteidl/plugins/kubeflow/common.proto";
 
+// Custom proto for torch elastic config for distributed training using 
+// https://github.com/kubeflow/training-operator/blob/master/pkg/apis/kubeflow.org/v1/pytorch_types.go
+message ElasticConfig {
```
fg91, line 12: FYI https://github.com/flyteorg/flyteidl/commit/6a7b31438821190be535cc6bbcb6f2939e5a056a

https://github.com/flyteorg/flyte/issues/3614

(Flytekit PR is not merged yet)

yubofredwang: @fg91 question, does elastic config require the master replica spec to be nil/not set?

fg91: Unfortunately Kubeflow is very poorly documented but from [their examples](https://github.com/kubeflow/training-operator/blob/master/examples/pytorch/elastic/imagenet/imagenet.yaml) this is my understanding, yes.

I also handled it like this in [this](https://github.com/flyteorg/flyteplugins/pull/343/files) PR in `flyteplugins` and it works.

pingsutw (approved):

yubofredwang:
> Looks great! My only nit is on proto location in respect to `v0` of the kubeflow operators. Not sure of the cleanest way to handle this (maybe this is it!) but it seems a little odd to have `v0` and `v1` in different locations.

The reason I put these into that specific location `kubeflow/plugins` was the following:
1. The three plugins proto share common parameters, which are specific to kubeflow plugins.
2. The new plugins now can keep the same name such as "DistributedMPITrainingTask". If I put it under same root directory, compiler will complain with conflicts since they are named the same.
3. I am kind of reluctant to rename the plugins to `pytorch_v1.proto` since later on after we deprecate v0, we might have to rename it again. Then at that time, we have to do another code change in both flytekit and flyteplugins repo

PR #387, opened 2023-03-31
Refactored literal type casting from flytepropeller

hamersaw:
# TL;DR
To allow re-use of literal casting in other repositories this PR refactors the code out of flytepropeller.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
^^^

## Tracking Issue
_NA_

## Follow-up issue
_NA_


clients/go/coreutils/casting.go
```diff
@@ -0,0 +1,367 @@
+package coreutils
```
yubofredwang, line 1: looks like this module was just providing type checking for upstream and downstream nodes. I think the naming of this file is a bit weird. or can you add more comments to explain what exactly this module does?

clients/go/coreutils/typing.go
```diff
@@ -0,0 +1,140 @@
+package coreutils
```
yubofredwang, line 1: Same question here, the naming of the file does not describe what this module is doing

PR #388, opened 2023-04-03
Add userid to security context

ByronHsu:
# TL;DR

Add execution userid to security context

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Tests

Did an e2e test with https://github.com/flyteorg/flyteadmin/pull/549
`user_identifier` was successfully injected to flyteCRD.
```
securityContext:
  run_as:
    user_identifier: byhsu
```

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

https://github.com/flyteorg/flyte/issues/3566


protos/flyteidl/core/security.proto
```diff
@@ -69,6 +69,9 @@ message Identity {
     // oauth2_client references an oauth2 client. Backend plugins can use this information to impersonate the client when
     // making external calls.
     OAuth2Client oauth2_client = 3;
+
+    // exec_user_id references the id of the user who executes the task.
+    string exec_user_id = 4;
```
wild-endeavor, line 74: @EngHabu what do you think about this?  We've had several requests over the past year and change to add some sort of human-level identity to the wf crd.  What do you think about this, just a free-form text?  presumably people can make this whatever they want, internal user id, email, jwt token possibly, encrypted encoded bytes, etc.

EngHabu: ic... trying to reason about this... any planned use for it in flyte components?

ByronHsu: in our case, we pass the `exec_user_id` to the flyteplugins because our plugins require this info

wild-endeavor: i think we should rename this field to just `identifier` or `user_identifier`.  @EngHabu ?

ByronHsu: `user_identifier` LGTM

wild-endeavor (approved):
can you update base and regenerate please?  also, before merging, can we have and test a sample middleware in the admin repo that will look for the `name` or `email` field in the JWT token, and fill this field in?  and let's confirm that we see it in the flyteworkflow CRD (`kubectl get fly <execid>`)  but in general looks good i think.

protos/flyteidl/core/security.proto
katrogan: How does this work for subworkflows and launch plans triggered by workflows? Will the user identifier be passed down or will flytepropeller's app credentials be passed instead?

wild-endeavor: Good point.  The audit trail should definitely encompass the fact that it came from propeller, but this field should continue to be the original user identifier string.  @ByronHsu this is possible with the design we're thinking right?  Or were you implying that this should be a list @katrogan ?

katrogan: I wasn't implying it should be a list but just wanted to understand the implications of hard-coding this to always reference a user if we anticipate app-driven identifiers

wild-endeavor: i'm okay adding additional fields as they become necessary.

ByronHsu:
@wild-endeavor sure i will test in [this pr](https://github.com/flyteorg/flyteadmin/pull/549/files) shortly

wild-endeavor:
did you ever get the time to test the admin change?

protos/flyteidl/core/security.proto
EngHabu: Just to confirm, we will pass this along when Propeller launches LPs from a Workflow Execution...
What will the value be for scheduled workflows? some system/default value?

EngHabu (approved):

ByronHsu:
@wild-endeavor e2e test done. Adding in the **Tests** section in the PR description. Thanks

wild-endeavor (approved):

eapolinario (approved):

eapolinario (approved):

PR #400, opened 2023-05-04
Add envs to execution spec 

pingsutw:
# TL;DR
Add envs to execution spec, allowing people to add env at execution time

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
^^^

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2447

## Follow-up issue
_NA_


protos/flyteidl/admin/execution.proto
```diff
@@ -310,6 +310,9 @@ message ExecutionSpec {
     // If enabled, all calculations are performed even if cached results would be available, overwriting the stored
     // data once execution finishes successfully.
     bool overwrite_cache = 22;
+
+    // Environment variables to be set for the execution.
+    Envs envs = 23;
```
hamersaw, line 315: What is the reason to use a separate struct for `Envs` rather than just setting this as:
```
map<string, string> environment_variables = 23;
```
It seems like just another layer of indirection?

protos/flyteidl/admin/common.proto
```diff
@@ -280,6 +280,14 @@ message Annotations {
     map<string, string> values = 1;
 }
 
+// Environment variable values to be applied to an execution resource.
+// In the future a mode (e.g. OVERRIDE, APPEND, etc) can be defined
+// to specify how to merge environment variables defined at registration and execution time.
+message Envs {
+    // Map of custom environment variables to be applied to the execution resource.
+    map<string, string> values = 1;
```
EngHabu, line 288: I don't have a good reason but we don't need to look up, we just need uniqueness. I think this should be a list.. i know it's a dict in python but it's a list in k8s...
We just have had too many "arrrrrgh, I wish we made it a list" moments in the past (variables in lps come to mind where maintaining original ordering is desired)

protos/flyteidl/admin/execution.proto
pingsutw: @EngHabu @hamersaw IIUC, the default value of map is {}. Adding another layer allows us to know value is set or unset. 
If people didn't add env to WorkflowExecutionConfig , the default value of env is always {} in it, and it may override the env in execution.

hamersaw: In the case of `interruptible` support here it was beneficial to know whether the flag was set or unset. In this case, I don't think it is necessary or has any benefit. A not provided `env` struct will be treated the same as one with 0 length right?

Per @EngHabu comment below. It sounds like we should list this anyways - I'm OK with this route. So we will need the struct `EnvironmentVariable` and then update the references to `repeated EnvironmentVariable env_vars = x;` or something like that.

pingsutw: we can't do something like below without adding another layer
```python
if override_env is not None:
    execution_env = override_env
```

wild-endeavor (approved):

hamersaw (approved):

PR #412, opened 2023-05-28
Add config override idl

ByronHsu:
# TL;DR

Added idl as discussed in https://github.com/flyteorg/flyte/pull/3553

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Tracking Issue
https://github.com/flyteorg/flyte/issues/475


protos/flyteidl/core/workflow.proto
```diff
@@ -249,4 +253,37 @@ message WorkflowTemplate {
 message TaskNodeOverrides {
     // A customizable interface to convey resources requested for a task container. 
     Resources resources = 1;
+
+    // Boolean that indicates if caching should be enabled
+    oneof cache_value {
```
hamersaw, line 258: For all of these `oneof`s we can also use `google.protobuf.BoolValue`, `google.protobuf.Int32Value`, etc which differentiate between being set and unset in addition to the values.

ByronHsu: Whats the concern of oneof?

katrogan (commented):
Since the fields you added to the TaskNodeOverrides are also used in matchable attributes [WorkflowExecutionConfig](https://github.com/flyteorg/flyteidl/blob/0dabef70811b92d63ac4abaff44eb0c96162b3c7/protos/flyteidl/admin/matchable_resource.proto#L105-L134), [ExecutionSpec](https://github.com/flyteorg/flyteidl/blob/0dabef70811b92d63ac4abaff44eb0c96162b3c7/protos/flyteidl/admin/execution.proto#L256-L321), [LaunchPlanSpec](https://github.com/flyteorg/flyteidl/blob/0dabef70811b92d63ac4abaff44eb0c96162b3c7/protos/flyteidl/admin/launch_plan.proto#L80-L136) and more, could we take this PR as an opportunity to encapsulate all these fields in a well-defined proto message (which we can then embed in `TaskNodeOverrides`) and in the future hopefully refactor all the existing specs to share as part of a separate GH issue? Currently adding a new execution-type field and understanding the pattern of overrides is onerous so it would be super helpful to establish a singular message/source of truth in this PR we could later migrate to elsewhere

PR #419, opened 2023-06-27
GPU Type

wild-endeavor:
# TL;DR
Adds support for passing along extra metadata related to resources (e.g. accelerator type) to the backend

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


hamersaw (approved):

pingsutw (approved):

protos/flyteidl/core/tasks.proto
```diff
@@ -61,6 +61,22 @@ message RuntimeMetadata {
...
+    string device = 1;
+    oneof partition_size_value {
+        bool unpartitioned = 2;
+        string partition_size = 3;
+    }
+}
+
+// Additional metadata associated with resources to allocate to a task
+message ResourceMetadata {
+    oneof accelerator_value {
```
EngHabu, line 75: Why oneof? if it's not set, it'll just be nil, do you want to allow nullifying a field? if so maybe that should go into the override message?

jeevb: The thought was to support other accelerator types like TPUs in the future. One option would have been to just add all possible accelerator types as separate fields without the guarantee that oneof provides - that only one type is set at any given point in time.

I actually did not consider nullifying an accelerator specified in the task metadata, in the override. That does sound nice. If we were to do that, would it be a bad idea to add a null_accelerator to the oneof?

EngHabu: Since oneofs do not change wire format, I would keep it out and kick the can down the road when we add a different accelerator to decide then.

There isn't a great way to nullify in Protos unfortunately...

jeevb: https://github.com/flyteorg/flyteidl/pull/442

PR #429, opened 2023-08-04
Add fields to specify ray head/worker resources + additional config map 

guozhen-la:

# TL;DR

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


protos/flyteidl/plugins/ray.proto
```diff
@@ -19,13 +21,19 @@ message RayCluster {
...
+  string k8s_service_account = 4;
 }
 
 // HeadGroupSpec are the spec for the head pod
 message HeadGroupSpec {
   // Optional. RayStartParams are the params of the start command: address, object-store-memory.
   // Refer to https://docs.ray.io/en/latest/ray-core/package-ref.html#ray-start
   map<string, string> ray_start_params = 1;
+  // Resource specification for ray head pod
+  core.Resources resources = 6;
```
pingsutw, line 36:
```suggestion
  core.Resources resources = 2;
```

protos/flyteidl/plugins/ray.proto
```diff
@@ -19,13 +21,19 @@ message RayCluster {
   HeadGroupSpec head_group_spec = 1;
   // WorkerGroupSpecs are the specs for the worker pods
   repeated WorkerGroupSpec worker_group_spec = 2;
+  // Namespace used to create the ray cluster
+  string namespace = 3;
```
pingsutw, line 25: To make it more generic, could we add it to the task metadata [here](https://github.com/flyteorg/flyteidl/blob/0dabef70811b92d63ac4abaff44eb0c96162b3c7/protos/flyteidl/core/tasks.proto#L65)

protos/flyteidl/plugins/ray.proto
```diff
@@ -19,13 +21,19 @@ message RayCluster {
...
+  // Kubernetes service account used by the ray cluster
+  string k8s_service_account = 4;
```
pingsutw, line 27: I just remembered that we could override the service account by using [pod template](https://docs.flyte.org/en/latest/deployment/configuration/general.html#compile-time-podtemplates). That said, we can remove `k8s_service_account` in the idl.

aybidi: @pingsutw how does it look like in the flytekit UX? we added `k8s_service_account` to the `RayJobConfig`. Would we not need it anymore and instead expect the user to override it using pod template?

aybidi: I think it may not be nice user experience if the user has to specify some attributes to override in `RayJobConfig` and some through a pod template

pingsutw: We want to make it general because others may want to override the service account for spark, Dask, TensorFlow tasks, etc. How about adding a namespace arg to the task decorator?
```python
@task(rayJobConfig(...), service_account="")
def ray_task()
```

aybidi: Are you suggesting just keeping the Ray-specific parameters in the RayJobConfig? Are we then going to create a parameter in the task decorator for each non-Ray configuration we want to override?

pingsutw: Namespace is not a common config that people will override so we can keep it in the rayjob. 
> Are you suggesting just keeping the Ray-specific parameters in the RayJobConfig? Are we then going to create a parameter in the task decorator for each non-Ray configuration we want to override?

Yes, ideally. people may want to override the service account for the spark job. any other configs you want to override?

aybidi: @pingsutw sorry we were away for a work event last week. So if I understand correctly, you're just suggesting moving `k8s_sa` outside of `rayjob` and allow users to configure it at the decorator level?

```python
@task(rayJobConfig(...), service_account="")
def ray_task():
    ...
```

aybidi: I understand the concern around generalizability (other tasks also configuring `service_account`). Is this still not a better user-experience to configure everything in the task config object versus a decorator-level argument?

```python
@task(rayJobConfig(namespace="", k8s="...",...))
def ray_task()

@task(sparkJobConfig(k8s="...",...))
def spark_task()
```

PR #432, opened 2023-08-17
Sort key and filters as a list of columns in list requests

iaroslav-ciupin:

# TL;DR
In order to fix SQL injection we will deprecate and restrict existing `sort_by` to just single column and introduce new field `sort_keys` that will accept array of columns with directions.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


protos/flyteidl/admin/common.proto
```diff
@@ -210,9 +210,13 @@ message ResourceListRequest {
     // +optional
     string filters = 4;
 
-    // Sort ordering.
+    // Sort column ordering.
     // +optional
-    Sort sort_by = 5;
+    // Deprecated.
```
katrogan, line 215: nit can we add a comment here to use sort_keys instead?

katrogan (approved):

hamersaw:
can we fix the generate verification test?

iaroslav-ciupin:
@hamersaw I'm not sure I know how to do this.

protos/flyteidl/admin/common.proto
```diff
@@ -210,9 +210,13 @@ message ResourceListRequest {
...
 
-    // Sort ordering.
+    // Sort column ordering.
     // +optional
-    Sort sort_by = 5;
+    // Deprecated. Use sort_keys instead.
+    Sort sort_by = 5 [deprecated = true];
+
+    // Sort column(s) with direction.
+    repeated Sort sort_keys = 6;
```
katrogan, line 219: does this work with http requests in grpc gateway v1? 
I recall we used unstructured string for filters (https://docs.flyte.org/en/latest/concepts/admin.html#adding-request-filters)

>  While this allows for a single implementation for both HTTP and gRPC, an important limitation is that fields mapped to the path pattern cannot be repeated and must have a primitive (non-message) type. Unfortunately this means that repeated string filters cannot use a proper protobuf message.

iaroslav-ciupin:
this change is not needed right now

PR #439, opened 2023-09-11
[flyteidl] Support attribute access on promises

ByronHsu:

# TL;DR

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


protos/flyteidl/core/types.proto
```diff
@@ -170,6 +170,26 @@ message OutputReference {
 
     // Variable name must refer to an output variable for the node.
     string var = 2;
+
+    repeated PromiseAtrribute attr_path = 3;
```
pingsutw, line 174: would it be better to name it `attr`? btw, could you add a comment?

pingsutw: should we just add an attribute and item in the `OutputReference`?
like
```proto
...
string attribute = 3;
string item = 4;
int index = 5;
```

ByronHsu: but attribute/item are list. Or you mean putting the above in `PromiseAtrribute`?
And in `OutputReference`, we still use 

```
repeated PromiseAtrribute
```

I think the idea to differentiate the attr to these 3 types is cleaner and more easier to understand

protos/flyteidl/core/types.proto
```diff
@@ -170,6 +170,26 @@ message OutputReference {
...
+// In the following example,
+// ```
+// @workflow
+// def wf():
+//     o = t1()
+//     t2(o.a["b"][0])
+// ```
+// the output reference t2 binds to has a list of PromiseAttribute ["a", "b", 0]
+
+message PromiseAtrribute {
```
wild-endeavor, line 188:
```suggestion
message PromiseAttribute {
```

protos/flyteidl/core/types.proto
ByronHsu: After the offline discussion. We will keep PromiseAttribute fields as is

pingsutw (approved):

wild-endeavor (approved):

ByronHsu:
Update 2023/10/14: 

After a discussion with @wild-endeavor, we would like to perform strong type check for fields in dataclasses. For example, 

```python
@dataclass
class Foo:
  a: int

@task
def t0() -> Foo:
  return Foo(a=1)

@task
def t1(x: str):
   ...

@workflow
def wf():
  f = to()
  t1(x=f.a)
```

In the original implementation, `wf` passed the compilation though `f.a` (`int`) and x (`str`) are different types. This is because we didn't store the type of dataclass fields in LiteralType, so we couldn't check them in the flytepropeller.

In order to do type check for dataclass field, I added `dataclass_type` in TypeStructure. It is a `map<str, LiteralType>`, which stores the mapping of dataclass field name to LiteralType. This also handles the case of nested dataclass because child dataclass can also be stored as a LiiteralType in the parent's `dataclass_type`.

Related Code Changes:
1. flyteidl PR: https://github.com/flyteorg/flyte/pull/4233
2. this commit in flytekit: https://github.com/flyteorg/flytekit/pull/1825/commits/722a9a6f1efd1611c45f436c7593a277ae3290ad
3. flytepropeller PR (depending on 1): https://github.com/flyteorg/flyte/pull/4232


I also conducted e2e test to ensure the behavior. In the following example, the workflow failed at compile time because of mismatched type.

```

@task
def t1() -> (List[str], Dict[str, str], foo):
    return ["a", "b"], {"a": "b"}, foo(a="b")


@task
def t6(a: int):
    return a


@workflow
def dataclass_type_check_fails_at_compile():
    l, d, f = t1()
    # f.a is str, but t6 expects int. This should fail at compile time
    t6(a=f.a)

```

PR #443, opened 2023-09-21
Add batched reasons to TaskExecutionEvent

andrewwdye:
# TL;DR
Deprecate `TaskExecutionEvent.reason` in favor of `TaskExecutionEvent.reasons` to send batched events from flytepropeller, each with their own timestamp.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
This change makes it possible to send kubernetes events as described in https://github.com/flyteorg/flyte/issues/3825 without overloading propeller or admin with a large increase in TaskExecutionEvents. It deprecates `TaskExecutionEvent.reason` in favor of a repeated `TaskExecutionEvent.reasons`, which includes separate timestamps for each event. This allows for sending multiple reasons as part of the same event.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3825

## Follow-up issue
N/A


protos/flyteidl/event/event.proto
```diff
@@ -154,6 +154,14 @@ message ParentNodeExecutionMetadata {
     string node_id = 1;
 }
 
+message BatchedReason {
```
katrogan, line 157: nit: seems confusing to call the individual reason message "Batched"

hamersaw: +1

protos/flyteidl/event/event.proto
```diff
@@ -154,6 +154,14 @@ message ParentNodeExecutionMetadata {
     string node_id = 1;
 }
 
+message Reason {
+    // An explanation for this event
+    string reason = 1;
```
katrogan, line 159: sorry to be so nitpicky - but won't this cause a stutter when accessing, e.g. `Reason.GetReason()` ?

either way don't feel too strongly

katrogan (approved):

protos/flyteidl/event/event.proto
andrewwdye: I've been waffling on this too. I didn't want to call this an `Event` because it's in an event. Maybe `EventReason.GetReason()` or `ReasonInfo.GetReason()`. Have a preference?

katrogan: both sound great! maybe slight preference for the former

hamersaw (approved):

katrogan (approved):

PR #450, opened 2023-09-29
Add driver and executor pod template fields to SparkJob

andrewwdye:
# TL;DR
Add driver and executor pod template fields to SparkJob

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
This change adds optional fields for pod and pod template name for the driver and executor pods. This mirrors the interface we have for container tasks in TaskTemplate and TaskMetadata.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/4105


protos/flyteidl/plugins/spark.proto
```diff
@@ -31,4 +33,24 @@ message SparkJob {
     // Domain name of your deployment. Use the form <account>.cloud.databricks.com.
     // This instance name can be set in either flytepropeller or flytekit.
     string databricksInstance = 9;
+    // The pod spec and metadata to be used as the base configuration when creating the driver Pod for this task.
+    // +optional
+    oneof driverPodValue {
```
katrogan, line 38: is this oneof to make it nillable?

andrewwdye: Correct

protos/flyteidl/plugins/spark.proto
```diff
@@ -31,4 +33,24 @@ message SparkJob {
...
+        core.K8sPod driverPod = 10;
```
wild-endeavor, line 39: Instead of using pod specs, should we just use a google.protobuf.Struct directly (basically just the `pod_spec` field in our `K8sPod` message?

Looking at what the [their guide](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/docs/user-guide.md#writing-driver-specification) is saying, it doesn't really look like a real K8s pod spec/template, which means we'd have to do massaging in the plugin before submitting it to the operator right?

It'd be nice if they just had a nice python class with all the right options, but it doesn't look like the `SparkConf` object is that.  So maybe just give users a nested hash map since that what it feels like that's all they provide.  Maybe one each for driver and executor.

andrewwdye: My thought in using the `core.K8sPod` here is to (1) pass labels and annotations as we do for container task [here](https://github.com/flyteorg/flytekit/blob/3cb79885049f9de6552d00bff1e9a57ba0d3ee0a/flytekit/core/container_task.py#L137-L141) and (2) reuse the [ToK8sPodSpec](https://github.com/flyteorg/flyteplugins/blob/6048689ca61f318ba30ac3e0df8729918314035e/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L258) in flyteplugins to merge various pod configs before extracting out the relevant ones to [build](https://github.com/flyteorg/flyteplugins/blob/24a2e436819d68ab7e1dd9cfcb2c2578ebed3e65/go/tasks/plugins/k8s/spark/spark.go#L126-L162) `sparkOp.DriverSpec` and `sparkOp.ExecutorSpec`.

wild-endeavor: oh ic.  is it always a subset though of pod spec?  or are there fields that users might want to set in the driver/executor specs that aren't in pod spec?

andrewwdye: Labels and annotations are not part of [PodSpec](https://pkg.go.dev/k8s.io/api/core/v1#PodSpec), but are [added to the SparkPodSpec](https://github.com/flyteorg/flyteplugins/blob/24a2e436819d68ab7e1dd9cfcb2c2578ebed3e65/go/tasks/plugins/k8s/spark/spark.go#L129-L130). Maybe we could pass the protobuf.Struct version of [PodTemplateSpec](https://pkg.go.dev/k8s.io/api/core/v1#PodTemplateSpec), which has both `PodSpec` and `ObjectMeta`, but I was err'ing on the side of consistency with [what we use for container task](https://github.com/flyteorg/flytekit/blob/3cb79885049f9de6552d00bff1e9a57ba0d3ee0a/flytekit/core/container_task.py#L137-L141). Happy to use an alternative here if more appropriate for Spark since we are just tunneling the info as long as we can send everything we need.

wild-endeavor: okay yeah that makes sense, apologies for distracting... we can always add more fields in the future if users ask for more.  I was initially a bit concerned that their pod spec isn't a real pod spec, even though they call it that (like the cpu limit is called a `CoreLimit`, among other differences), but yeah we can easily add to the mapping as users request them in the block where we're already doing the construction of their podspec object.  thank you!

protos/flyteidl/plugins/spark.proto
```diff
@@ -31,4 +33,24 @@ message SparkJob {
...
+    // The pod spec and metadata to be used as the base configuration when creating the executor Pods for this task.
+    // +optional
+    oneof executorPodValue {
+        core.K8sPod executorPod = 12;
+    }
+    // Reference to an existing PodTemplate k8s resource to be used as the base configuration when creating the 
+    // executor Pods for this task. If this value is set, the specified PodTemplate will be used instead of, but applied
+    // identically as, the default PodTemplate configured in FlytePropeller.
+    // +optional
+    string executorPodTemplateName = 13;
```
hamersaw, line 55: Is there any reason these can't just be the PodTemplate / PodTemplateName from the TaskExecutionContext? IIRC this is how we built some of the other plugins (ie. Dask, KF operator).

andrewwdye: Do you mean use the same pair for both executor and driver, use the existing pair from TaskExecutionContext for one of these and plumb the other here, or something else?

andrewwdye:
Closing in favor of https://github.com/flyteorg/flyte/pull/4179

protos/flyteidl/plugins/spark.proto
hamersaw: Either or. Both of these are available in the task decorator in flytekit, so if we added them both at the `spark_config` level -- ex:
```
@task(
    task_config=Spark(
        spark_config=[
            executor_pod_template=...
            driver_pod_template=...
        ]
    )
    pod_template=odTemplate(
        primary_container_name="primary",
        labels={"lKeyA": "lValA", "lKeyB": "lValB"},
        annotations={"aKeyA": "aValA", "aKeyB": "aValB"},
        pod_spec=V1PodSpec(
            volumes=[V1Volume(name="volume")],
            tolerations=[
                V1Toleration(
                    key="num-gpus",
                    operator="Equal",
                    value=1,
                    effect="NoSchedule",
                ),
            ],
        )
)
```
it doesn't make sense to me intuitively how they would be applied. For things like resources / etc that are already available in the task decorator we implemented (in other plugins) that if set at the task-level they would be applied to all replicas (ie. executor / driver) unless there was a specific override there. So IMO we should apply task-level to the executor / driver by default and put an override on the driver only.

andrewwdye: I waffled on this a bit since the spark on k8s op itself has moved to an abstraction of explicit configs for each driver and executor, but yeah I tend to agree here that utilizing the top level `TaskTemplate.PodTemplate` as the primary override for both + a driver-specific `PodTemplate` in the SparkJob is more consistent with flytekit and other plugins. 

We'll need to clarify the order of precedence w.r.t. top level configs (i.e., TaskTemplate.Container), top level pod template (i.e., TaskTemplate.PodTemplate.PrimaryContainerName), and driver-specifc config (i.e., SparkJob.PodTemplate.PrimaryContainerName). The same set of clarifications are needed for other configs (i.e., memory/cores, which can also be specified as spark config -- `spark.executor.requests.memory`).

So the proposed python devx would be
```
@task(
    task_config=Spark(
        spark_conf={
           "spark.driver.memory": "4G",
           "spark.executor.memory": "8G",
        },
        # driver only pod template overrides (takes precedence over top level pod_template)
        driver_pod_template=...
    ),
    # default pod template overrides for BOTH driver and executor
    pod_template=odTemplate(
        primary_container_name="primary",
        labels={"lKeyA": "lValA", "lKeyB": "lValB"},
        annotations={"aKeyA": "aValA", "aKeyB": "aValB"},
        pod_spec=V1PodSpec(
            volumes=[V1Volume(name="volume")],
            tolerations=[
                V1Toleration(
                    key="num-gpus",
                    operator="Equal",
                    value=1,
                    effect="NoSchedule",
                ),
            ],
        )
    ),
)
```

@pingsutw @wild-endeavor any thoughts?

pingsutw: I agree it's very confusing to have `podTemplate` in both `@task` and `spark` config, but I still think we need `worker_pod_template` in spark config. 

People may only want to override the `worker_pod_template`, for example, add gpu selector, or something else.

wild-endeavor: i'd personally rather have the original where both pod templates are in the spark config, and then just error if the pod level config is specified.  but i'm okay any way.  I just feel like applying one could make equally convincing arguments that the normal top level config should be for executors or for drivers.

andrewwdye: Thanks for the thoughtful feedback. After weighing this + the suggestions in the [migrated PR](https://github.com/flyteorg/flyte/pull/4179#pullrequestreview-1662713211), I'd like to propose
* A RoleSpec definition in common.proto
* Inclusion of driver and executor role specs in SparkJob
* Error in spark plugin if TaskTemplate.PodTemplate is set
* Use this pattern as a deprecation path for TaskTemplate.PodTemplate for Dask, MPI, PyTorch, and Ray should we choose to have per-role PodTemplates/overrides (i.e., start erroring if both TaskTemplate.PodTemplate and worker RoleSpec are specified

Will summarize fully in https://github.com/flyteorg/flyte/pull/4179 and try to move the remainder of the communication there

